PR #5, opened 2022-11-02
fe and cn use configmap

shileifu:


pkg/common/resource_utils/configmap.go
```diff
@@ -0,0 +1,151 @@
...
+//getPort get ports from config file.
+func GetPort(config map[string]interface{}, key string) int32 {
+	if key == HTTP_PORT {
+		if v, ok := config[HTTP_PORT]; ok {
+			if port, err := strconv.ParseInt(v.(string), 10, 32); err == nil {
+				return int32(port)
+			}
+		}
+
+		return 8030
```
DorianZheng, line 68: magic numbers

script/fe_prestop.sh
```diff
@@ -19,7 +19,8 @@ SELF_HOST=
 #left_node_lists node lists of except self.
 LEFT_NODE_LSITS=
 PROBE_LEADER_PODX_TIMEOUT=120 # at most 60 attempts
-STARROCK_HOME=${STARROCK_HOME:-"/opt/starrocks"}
+STARROCKS_ROOT=${STARROCKS_HOME:-"/opt/starrocks"}
+STARROCKS_HOME=$STARROCKS_ROOT/fe
```
DorianZheng, line 23: `STARROCKS_FE_HOME`

script/cn_stop.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+STARROCKS_ROOT=${STARROCKS_ROOT:-"/opt/starrocks"}
+STARROCKS_HOME=$STARROCKS_ROOT/be
```
DorianZheng, line 4: `STARROCKS_CN_HOME`

pkg/fe_controller/fe_pod.go
```diff
@@ -24,6 +24,13 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+const (
+	meta_path = "/opt/starrocks/fe/meta"
```
DorianZheng, line 28: Is it a standard option to deploy under `/opt/`

pkg/cn_controller/cn_pod.go
```diff
@@ -22,6 +22,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"strconv"
+)
+
+const (
+	log_path = "/opt/starrocks/be/log"
```
DorianZheng, line 29: is it better to give `cn` a different path than `be`?

script/fe_prestop.sh
shileifu: this is starrocks script set, we follow the starrocks.

pkg/fe_controller/fe_pod.go
shileifu: the path is typical for container deployment, it is not absolute we can use a different path.

pkg/cn_controller/cn_pod.go
shileifu: the log path is starrocks be set, operator is only follow the rule.

kevincai: it is possible to set to `cn/log`

in dockerfile, copy `output/be` to `/opt/starrocks/cn` , set STARROCKS_ROOT=/opt/starrocks, set STARROCKS_HOME=${STARROCKS_ROOT}/cn

pkg/cn_controller/cn_controller.go
```diff
@@ -73,17 +75,30 @@ func (cc *CnController) Sync(ctx context.Context, src *srapi.StarRocksCluster) e
 		return nil
 	}
 
+	//get the cn configMap for resolve ports.
+	//2. get config for generate statefulset and service.
```
kevincai, line 79: get config **to** generate ...

pkg/cn_controller/cn_controller.go
```diff
@@ -73,17 +75,30 @@ func (cc *CnController) Sync(ctx context.Context, src *srapi.StarRocksCluster) e
...
+	config, err := cc.GetCnConfig(ctx, &cnSpec.ConfigMapInfo, src.Namespace)
+	if err != nil {
+		klog.Error("CnController Sync ", "resolve cn configmap failed, namespace ", src.Namespace, " configmapName ", cnSpec.ConfigMapInfo.ConfigMapName, " configMapKey ", cnSpec.ConfigMapInfo.ResolveKey, " error ", err)
+		return err
+	}
+
+	feconfig, _ := cc.getFeConfig(ctx, &src.Spec.StarRocksFeSpec.ConfigMapInfo, src.Namespace)
+	//merge fe config for simplify processing.
+	for k, v := range feconfig {
+		config[k] = v
```
kevincai, line 89: it is possible that fe config and cn config has the same config key, this merge operation requires that the two components can't not have same key name, otherwise one will overwrite the other.

Doesn't sound reasonable just to add new restrictions to components configuration because of this operational detailed implementation thing.

pkg/common/resource_utils/configmap.go
```diff
@@ -0,0 +1,151 @@
...
+	klog.Info("the resolve message ", value)
+	viper.SetConfigType("properties")
+	viper.ReadConfig(bytes.NewBuffer([]byte(value)))
+
+	return viper.AllSettings(), nil
+}
+
+//getPort get ports from config file.
+func GetPort(config map[string]interface{}, key string) int32 {
+	if key == HTTP_PORT {
```
kevincai, line 61:
```
defaultPort := new map[string][int]{HTTP_PORT: 8030, xxx: 9030}

if v, ok := config[key]; ok {
    ...
}
return defaultPort[key]
```

script/cn_stop.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
kevincai, line 1: rename to cn_prestop.sh

PR #30, opened 2023-01-18
update liveness timeout

shileifu:


kevincai (approved):

pkg/be_controller/be_pod.go
```diff
@@ -177,8 +177,7 @@ func (be *BeController) buildPodTemplate(src *v1alpha12.StarRocksCluster, beconf
 			}}},
 		},
 		LivenessProbe: &corev1.Probe{
-			FailureThreshold: 60,
```
dengliu, line 180: I understand there is a default value for `FailureThreshold` .
however, would it be better to set it explicitly for better readability?

pkg/cn_controller/cn_pod.go
```diff
@@ -150,8 +150,7 @@ func (cc *CnController) buildPodTemplate(src *v1alpha12.StarRocksCluster, cnconf
 			}}},
 		},
 		LivenessProbe: &corev1.Probe{
-			FailureThreshold: 60,
```
dengliu, line 153: ditto

pkg/fe_controller/fe_pod.go
```diff
@@ -194,8 +194,7 @@ func (fc *FeController) buildPodTemplate(src *v1alpha12.StarRocksCluster, feconf
 			}}},
 		},
 		LivenessProbe: &corev1.Probe{
-			FailureThreshold: 60,
```
dengliu, line 197: ditto

pkg/be_controller/be_pod.go
shileifu: OK

kevincai (approved):

dengliu (approved):

PR #45, opened 2023-02-14
fix be pod error log-storage-name and annotations

dxheming:
fix be pod error log-storage-name and annotations

Signed-off-by: haiming <dxheming@163.com>


shileifu (approved):

kevincai (approved):

kevincai:
thanks for the fix!

PR #90, opened 2023-04-15
Add ARM64 docker support

zorrofox:
Update the docker file for ARM64 platform support.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ COPY . .
 RUN make test
 
 # Build the binary
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags="${LDFLAGS:-}" -o /app/sroperator cmd/main.go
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -13,7 +13,7 @@ COPY . .
 RUN make test
 
 # Build the binary
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags="${LDFLAGS:-}" -o /app/sroperator cmd/main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=${BUILDARCH} go build -ldflags="${LDFLAGS:-}" -o /app/sroperator cmd/main.go
```
kevincai, line 16: is this BUILDARCH an env var or a build arg?

zorrofox: Yes, please reference this link:https://docs.docker.com/build/building/multi-platform/

kevincai:
done in #135

PR #102, opened 2023-04-27
Update starrocks-fe-and-cn-be.yaml

alberttwong:
using scarf for download.  
Signed-off-by: Albert Wong <atwong@alumni.uci.edu>


kevincai:
not necessary any more.

PR #129, opened 2023-06-15
[feature] support mount secrets in operator and helm chart

yandongxiao:
# Requirement

Currently, in the SR cluster, there are three types of components available: BE, FE, and CN. These components are already capable of mounting a ConfigMap object. The current requirement is for the StarRocks Operator to support mounting Secret objects when creating an SR cluster.

Specifically:

1. Mounting Secret objects is an optional choice when creating an SR cluster with the StarRocks Operator.
2. The StarRocks Operator does not need to create Secret objects.
3. The StarRocks Operator mounts Secret objects based on the CRD Spec.

# Design

## Operator

Add the `secrets []SecretInfo` field to StarRocksFeSpec, StarRocksBeSpec, and StarRocksCnSpec to support mounting multiple Secret objects.

The definition of the SecretInfo type is as follows:

```go
type SecretInfo struct {
    Name      string `json:"name"`
    MountPath string `json:"mountPath"`
}
```

Here are two CR definitions that mount Secret objects:

```yaml
apiVersion: starrocks.com/v1
kind: StarRocksCluster
metadata:
  name: starrockscluster-sample
  namespace: starrocks
spec:
  starRocksFeSpec:
    image: starrocks/fe-ubuntu:2.5.4
    replicas: 1
    requests:
      cpu: 1
      memory: 1Gi
    secrets:
      - name: foo
        mountPath: /data/m1
      - name: bar
        mountPath: /data/m2
  starRocksBeSpec:
    image: starrocks/be-ubuntu:2.5.4
    replicas: 1
    requests:
      cpu: 1
      memory: 1Gi
    secrets:
      - name: foo
        mountPath: /data/m1
      - name: bar
        mountPath: /data/m2
```

## helm chart

When deploying an SR cluster using Helm Chart, it relies on the StarRocksCluster CRD. Therefore, the changes to the Helm Chart include:

1. Creating corresponding Secret objects based on the existence of Values.secrets.
2. Mounting the relevant secret information based on the existence of Values.starrocksBeSpec.secrets.

```yaml
# define secrets in values.yaml
secrets:
  - name: secret1
    data:
      key1: value1
      key2: value2
  - name: secret2
    data:
      key1: value3
      key2: value4

# mount secrets in template
  secrets:
    - name: secret1
      mountPath: /etc/secret1
    - name: secret2
      mountPath: /etc/secret2
```


helm-charts/charts/kube-starrocks/values.yaml
kevincai, line 1: share add hints to the secrets configuration in this values.yaml, allow chart user to know how to add secrets if necessary.

yandongxiao: good suggestion !

kevincai (approved):

PR #136, opened 2023-06-26
support customized pod scheduler

yandongxiao:


pkg/sub_controller/be/be_pod.go
```diff
@@ -190,6 +190,7 @@ func (be *BeController) buildPodTemplate(src *srapi.StarRocksCluster, beconfig m
 		ImagePullSecrets:              beSpec.ImagePullSecrets,
 		NodeSelector:                  beSpec.NodeSelector,
 		HostAliases:                   beSpec.HostAliases,
+		SchedulerName:                 beSpec.SchedulerName,
```
kevincai, line 193: what if the `SchedulerName` is not present in CRD definition, will this be still fine to provide an empty scheduler?

kevincai (commented):
need change on helm charts as well.

pkg/sub_controller/be/be_pod.go
yandongxiao: If `beSpec.SchedulerName` is not assigned, then `beSpec.SchedulerName` is a zero value, which is an empty string for string type. `SchedulerName` is a string type, as a field in the `PodSpec` struct, and is also an empty string by default.

PR #137, opened 2023-06-26
[feature] support customized pod scheduler

yandongxiao:


pkg/sub_controller/be/be_pod.go
```diff
@@ -190,6 +190,7 @@ func (be *BeController) buildPodTemplate(src *srapi.StarRocksCluster, beconfig m
 		ImagePullSecrets:              beSpec.ImagePullSecrets,
 		NodeSelector:                  beSpec.NodeSelector,
 		HostAliases:                   beSpec.HostAliases,
+		SchedulerName:                 beSpec.SchedulerName,
```
yandongxiao, line 193: If `beSpec.SchedulerName` is not assigned, then `beSpec.SchedulerName` is a zero value, which is an empty string for string type. `SchedulerName `is a string type, as a field in the `PodSpec` struct, and is also an empty string by default.

kevincai (commented):
please update helm chart templates as well.

kevincai (approved):

PR #141, opened 2023-06-28
clean up helm chart values.yaml

yandongxiao:


helm-charts/charts/kube-starrocks/values.yaml
```diff
@@ -28,113 +31,98 @@ starrocksOperator:
...
   fsGroup: 0
-
   # specify the service name and port config and serviceType
   # the service type refer https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types
   service:
     # the fe service type, only supported ClusterIP, NodePort, LoadBalancer, ExternalName
     type: "ClusterIP"
     # the loadBalancerIP for static ip config when the type=LoadBalancer and loadbalancerIp is not empty.
-    loadbalancerIP: ""
+    loadBalancerIP: ""
```
kevincai, line 65: this will cause backwards compatible issue. if user already configured with `loadbalancerIP`, after upgrade, the configure will be lost, right?

yandongxiao: ok, I will cancel the modification.

helm-charts/charts/kube-starrocks/values.yaml
```diff
@@ -1,18 +1,21 @@
...
+  # the v1alpha1 version is also supported.
+  # supported value: v1, v1alpha1.
   crdVersion: v1
 
 # the starrockscluster crd' name.
 nameOverride: ""
+
 starrocksOperator:
-  # provide a name for operator deployment.
   enabled: true
```
kevincai, line 19: can add comments to these fields that are not obvious to end user.

helm-charts/charts/kube-starrocks/values.yaml
```diff
@@ -28,36 +31,31 @@ starrocksOperator:
...
-
 # deploy a starrocks cluster
 starrocksCluster:
-  # Provide a name for starprocks cluster
+  # the name of starrockscluster cluster, if not set, the chart name will be used.
   name: ""
+  # the namespace of starrockscluster cluster, if not set, the release namespace will be used.
   namespace: ""
-  # annotations for starrocks cluster
+  # annotations for starrocks cluster.
```
kevincai, line 40: this annotations is used or not?

helm-charts/charts/kube-starrocks/values.yaml
```diff
@@ -28,36 +31,31 @@ starrocksOperator:
...
-  # Number of replicas to deploy for a fe statefulset.
+  # number of replicas to deploy for a fe statefulset.
   replicas: 1
   image:
     # image sliced by "repository:tag"
     repository: starrocks/fe-ubuntu
     tag: 2.5.5
-
-  # add annotations for fe pods. example, if you want to config monitor for datadog, you can config the annotations. etc...
+  # add annotations for fe pods. example, if you want to config monitor for datadog, you can config the annotations.
```
kevincai, line 53:
```suggestion
  # add annotations for fe pods. For example, if you want to config monitor for datadog, you can config the annotations.
```

helm-charts/charts/kube-starrocks/values.yaml
yandongxiao: It is used in helm chart template to add annotation for starrockscluster

kevincai: but in starrockscluster,  operator doesn't use it at all?

helm-charts/charts/kube-starrocks/values.yaml
yandongxiao: the problem is the comment is not correct for `enabled` filed. And It is used together with `global.rbac.create`, I fix the comment there.

helm-charts/charts/kube-starrocks/values.yaml
yandongxiao: operator will use it to do restart operation. for exmaple: 
```
if !be.statefulsetNeedRolloutRestart(src.Annotations, est.Annotations) && ok {
	// when restart we add `AnnotationRestart` to annotation. so we should add again when we equal the exsit statefulset and new statefulset.
	anno := rutils.Annotations{}
	anno.Add(common.KubectlRestartAnnotationKey, est.Spec.Template.Annotations[common.KubectlRestartAnnotationKey])
	new.Spec.Template.Annotations = anno
}
```

helm-charts/charts/kube-starrocks/values.yaml
kevincai: ok

helm-charts/charts/kube-starrocks/values.yaml
kevincai: ok, may be more examples to be added here. sounds like only some specific annotations are acceptable here?

yandongxiao: good suggestion! maybe, I will check it, and add comment in detail.

helm-charts/charts/kube-starrocks/values.yaml
```diff
@@ -67,74 +65,64 @@ starrocksFESpec:
...
   serviceAccount: ""
-
   # If specified, the pod's nodeSelector，displayName="Map of nodeSelectors to match when scheduling pods on nodes"
   # Ref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector
   nodeSelector: {}
-  #  beta.kubernetes.io/arch: amd64
-  #  beta.kubernetes.io/os: linux
-
+    # beta.kubernetes.io/arch: amd64
+    # beta.kubernetes.io/os: linux
```
kevincai, line 77:
```suggestion
    # kubernetes.io/arch: amd64
    # kubernetes.io/os: linux
```

should be no `beta` prefix?

helm-charts/charts/kube-starrocks/values.yaml
```diff
@@ -67,74 +65,64 @@ starrocksFESpec:
...
-  #
-  # e.g. secret environment variable:
-  #  - name: USERNAME
-  #    valueFrom:
-  #      secretKeyRef:
-  #        name: mysecret
-  #        key: username
-
-  # Pod affinity
+    # e.g. static environment variable:
```
kevincai, line 92: we can add a kubernetes doc link here, here only lists two types of env var, but kubernetes supports more than that, we can lead the user to check the doc on how to set the env var themselves.

helm-charts/charts/kube-starrocks/values.yaml
yandongxiao: yes, no beta prefix. And there are three places that should be fixed like this. I will change it myself.

yandongxiao: enn, my local cluster build by kind has both labels（kubectl get nodes -oyaml）, like this:
```
beta.kubernetes.io/arch: arm64
beta.kubernetes.io/os: linux
kubernetes.io/arch: arm64
kubernetes.io/os: linux
```
whatever, I remove the `beta` prefix.

helm-charts/charts/kube-starrocks/values.yaml
yandongxiao: good suggestion, I will fix it.

kevincai (approved):

PR #156, opened 2023-07-03
[feature] support volumes mounting for configmaps

yandongxiao:


pkg/k8sutils/templates/pod/spec.go
```diff
@@ -78,8 +78,29 @@ func getProbe(port int32, path string) corev1.ProbeHandler {
 	}
 }
 
+func MountConfigMaps(volumes []corev1.Volume, volumeMounts []corev1.VolumeMount,
+	configMaps []v1.ConfigMapReference) ([]corev1.Volume, []corev1.VolumeMount) {
+	for _, configMap := range configMaps {
+		volumes = append(volumes, corev1.Volume{
+			Name: configMap.Name,
```
kevincai, line 85: whether need to marshall or randomize this volume name, if the configmap name conflicts some other volume name (e.g. user specify a configmap named fe-meta)

config/crd/bases/starrocks.com_starrocksclusters.yaml
```diff
@@ -4104,6 +4135,20 @@ spec:
...
+                    description: the reference for configMap which allow users to
+                      mount any files to container.
+                    items:
+                      properties:
+                        mountPath:
+                          description: Path within the container at which the volume
+                            should be mounted.  Must not contain ':'.
+                          type: string
+                        name:
+                          description: This must match the Name of a Volume.
```
kevincai, line 4148: looks like this description is not accurate, it is the configmap's name, and the mounted volume name is an internal name, is transparent to the user.

kevincai (approved):

PR #157, opened 2023-07-03
Feature/ensure pods restart when configmap changes

yandongxiao:


helm-charts/charts/kube-starrocks/templates/_helpers.tpl
```diff
@@ -120,3 +120,31 @@ be.conf: |
...
+  {{- printf "%s" $hash }}
+{{- end }}
+
+
+{{/*
+starrockscluster.be.config.hash is used to calculate the hash value of the be.conf, and due to the length limit, only
+the first 8 digits are taken, which will be used as the annotations for pods.
+*/}}
+{{- define "starrockscluster.be.config.hash" }}
+  {{- $hash := toJson .Values.starrocksBeSpec.config | sha256sum | trunc 8 }}
```
kevincai, line 139: if only use the first 8 digits, how about change to crc32?

yandongxiao: helm may not support crc32, see http://masterminds.github.io/sprig/

kevincai: ok

helm-charts/charts/kube-starrocks/templates/_helpers.tpl
```diff
@@ -120,3 +120,31 @@ be.conf: |
 {{- define "starrockscluster.be.log.path" -}}
 {{- print "/opt/starrocks/be/log" }}
 {{- end }}
+
+{{/*
+starrockscluster.fe.config.hash is used to calculate the hash value of the fe.conf, and due to the length limit, only
+the first 8 digits are taken, which will be used as the annotations for pods.
+*/}}
+{{- define "starrockscluster.fe.config.hash" }}
+  {{- $hash := toJson .Values.starrocksFESpec.config | sha256sum | trunc 8 }}
```
kevincai, line 129: what if user provided an empty fe/be/cn config? will this be still ok?

kevincai (approved):

PR #165, opened 2023-07-07
Feature/support to config toleration node selector annotation for operator pod

yandongxiao:


helm-charts/charts/kube-starrocks/templates/starrocks-operator/namespace.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
```
kevincai, line 1: I am thinking that the namespace should be respect of `helm -n` option, if we specify namespace in template, will it still respect the `-n <namespace>` issued by `helm`?

helm-charts/charts/kube-starrocks/templates/starrocks-operator/deployment.yaml
```diff
@@ -46,6 +54,15 @@ spec:
         # TODO(user): Configure the resources accordingly based on the project requirements.
         # More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
         resources:
-{{ toYaml .Values.starrocksOperator.resources | indent 10 }}
+          {{- toYaml .Values.starrocksOperator.resources | nindent 10 }}
       serviceAccountName: {{ template "kube-starrocks.operator.serviceAccountName" . }}
       terminationGracePeriodSeconds: 10
+      {{- if .Values.starrocksOperator.nodeSelector }}
+      nodeSelector:
+        {{- toYaml .Values.starrocksOperator.nodeSelector | nindent 8 }}
```
kevincai, line 62: is there any guideline or code style that those control statement should be written with no indention and those `toYaml` content output is controlled by `nindent`?

helm-charts/charts/kube-starrocks/templates/starrocks-operator/namespace.yaml
yandongxiao: There are a few things that need to be clarified:

1. By default, the Operator is deployed to `~/.kube/config` in the Namespace specified by `context.namespace`.
2. If you execute `helm install -n <namespace>`, the Operator will be installed in that Namespace.
3. If `.Values.starrocksOperator.namespaceOverride` is specified, it takes the highest priority.

Creating helm-charts/charts/kube-starrocks/templates/starrocks-operator/namespace.yaml ensures the existence of the namespace.

helm-charts/charts/kube-starrocks/templates/starrocks-operator/deployment.yaml
yandongxiao: If you  execute `helm create foo`, a helm chart demo will be created, and I follow this code style. e.g. 

```
      {{- with .Values.podAnnotations }}
      annotations:
        {{- toYaml . | nindent 8 }}
      {{- end }}

```

helm-charts/charts/kube-starrocks/templates/starrocks-operator/namespace.yaml
kevincai: ok. remember to add cases to cover this scenario, ensure when user specify `-n <namespace>`. it will take effect.

if `.Values.starrocksOperator.namespaceOverride`, our operator will be deployed to `default` namespace?

kevincai (approved):

PR #172, opened 2023-07-12
[bugfix] fix some bugs

yandongxiao:
1. reuse api.md and its script, because it is linked in StarRocks Doc
2. delete namespace.yaml, otherwise helm install will failed.
3. use resources not resource in values.yaml


helm-charts/charts/kube-starrocks/values.yaml
```diff
@@ -274,7 +274,7 @@ starrocksCnSpec:
     #     scaleDown:
     #       selectPolicy: Disabled
   # define resources requests and limits for cn pods.
-  resource:
+  resources:
```
kevincai, line 277: could cause compatible issue when the user upgrade from previous chart version to this, without updating its values.yaml?

yandongxiao: yes, This is a problem. But this misspell should be fixed.
So, we both support resources and resource?

kevincai: yes, please. recommends the right spelling, but can still work if mis-spelled for now, possibly generating a warning message? can possibly remove the support after 2-3 releases in future.

helm-charts/charts/kube-starrocks/templates/starrocks/starrockscluster.yaml
```diff
@@ -14,9 +14,11 @@ spec:
   starRocksFeSpec:
     image: "{{ .Values.starrocksFESpec.image.repository }}:{{ .Values.starrocksFESpec.image.tag }}"
     replicas: {{ .Values.starrocksFESpec.replicas }}
-{{- if .Values.starrocksFESpec.resources }}
-{{ toYaml .Values.starrocksFESpec.resources | indent 4 }}
-{{- end }}
+    {{- if .Values.starrocksFESpec.resources }}
+    {{- toYaml .Values.starrocksFESpec.resources | nindent 4 }}
+    {{- else if .Values.starrocksFESpec.resource }}
```
kevincai, line 19: does helm chart template support comments? shall add comments here to explain why if-else if here.

kevincai (approved):

PR #173, opened 2023-07-12
Feature/add test cases

yandongxiao:


helm-charts/charts/index.yaml
```diff
@@ -30,4 +30,572 @@ entries:
     urls:
     - https://github.com/StarRocks/starrocks-kubernetes-operator/releases/download/v1.7.0/kube-starrocks-chart-1.7.0.tgz
     version: 1.7.0
-generated: "2023-07-12T17:12:54.238881+08:00"
+  - apiVersion: v2
```
kevincai, line 33: we may want to remove a few historic release version, since the corresponding fe/be image are already removed, the chart version will not work anyway.

```
error scanning image starrocks/fe-ubuntu:2.5.0-fix-uid: image not found (package kube-starrocks:1.4.2)
```

helm-charts/charts/index.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - starrocks
+    - database
+    - olap
+    kubeVersion: '>=1.18.3-0'
+    maintainers:
+    - email: caixiaohua@starrocks.com
+      name: Kevin Cai
+    - email: shensida@starrocks.com
+      name: SidaShen
+    - email: fushilei@starrocks.com
```
kevincai, line 24: delete this contact since 1.7.0, add yourself into the list.

kevincai (approved):

PR #189, opened 2023-07-24
update some files for release v1.7.1

yandongxiao:


scripts/artifacts.sh
```diff
@@ -1,5 +1,8 @@
 #! /bin/sh
 
+set -e
+set -x
```
kevincai, line 4:
```suggestion
set -ex
```

kevincai (approved):

PR #193, opened 2023-07-25
create namespace for starrocks cluster

yandongxiao:


helm-charts/charts/kube-starrocks/values.yaml
```diff
@@ -51,8 +51,21 @@ starrocksOperator:
 starrocksCluster:
   # the name of starrockscluster cluster, if not set, the chart name will be used.
   name: ""
-  # the namespace of starrockscluster cluster, if not set, the release namespace will be used.
-  namespace: ""
+  # By default, starrockscluster will be deployed in the same namespace as the
+  # operator, and normally you can use helm --namespace to specify the namespace.
+  # This is used to create a different namespace for starrockscluster cluster with specified annotations.
+  namespace:
```
kevincai, line 57: what if the user uses an old version of values.yaml which gives starrocksCluster.namespace != "", will it report error when upgrades to this new version?

helm-charts/charts/kube-starrocks/values.yaml
```diff
@@ -51,8 +51,21 @@ starrocksOperator:
...
+    # if not set, the release namespace will be used.
+    # if set, the namespace will be created, and starrocks cluster and operator
+    # will be deployed in different namespaces.
+    name: ""
+    # annotations for starrockscluster namespace.
+    annotations: {}
```
kevincai, line 63: or rather than asking user to provide annotations here, can we have a `runOnOpenShift` option, if user sets it to `true`, chart automatically generate the annotation for the namespace.

Again, if the namespace already exists, will this change cause upgrade failure? or what exactly the behavior is?

helm-charts/charts/kube-starrocks/values.yaml
yandongxiao: yes, this will be a break change. And user need to update its values.yaml file

PR #204, opened 2023-08-02
support fe proxy

yandongxiao:
- use subchart to support multiple starrocks clusters
- support fe proxy


helm-charts/charts/kube-starrocks/charts/starrocks/templates/fe-proxy/feproxy.yaml
```diff
@@ -0,0 +1,146 @@
...
+      labels:
+        {{- include "feproxy.labels" . | nindent 8 }}
+    spec:
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 101
+        runAsGroup: 101
+      {{- if .Values.feProxy.nodeSelector }}
+      nodeSelector:
+        {{- toYaml .Values.feProxy.nodeSelector | nindent 8 }}
```
kevincai, line 27: probably need provide `affinity` and `tolerations` related configuration to end user as well.

helm-charts/charts/kube-starrocks/charts/starrocks/templates/fe-proxy/feproxy.yaml
```diff
@@ -0,0 +1,146 @@
+{{- if .Values.feProxy.enabled }}
+
+# create fe proxy deployment
+apiVersion: apps/v1
+kind: Deployment
```
kevincai, line 5: expect this deployment can be managed by operator, so it is part of the crd.

PR #211, opened 2023-08-08
add fe proxy

yandongxiao:


helm-charts/charts/kube-starrocks/charts/starrocks/values.yaml
```diff
@@ -408,3 +408,54 @@ configMaps: []
...
+    tag: 1.24.0
+  resources:
+    limits:
+      cpu: 1
+      memory: 2Gi
+    requests:
+      cpu: 1
+      memory: 2Gi
+  # set the resolver for nginx server
+  resolver: kube-dns.kube-system.svc.cluster.local
```
kevincai, line 428: can we make this default?

helm-charts/charts/kube-starrocks/charts/starrocks/values.yaml
```diff
@@ -408,3 +408,54 @@ configMaps: []
   #     key: |
   #       this is the content of the configmap
   #       when mounted, key will be the name of the file
+
+# specify the fe proxy deploy or not.
+feProxy:
+  # specify the fe proxy deploy or not.
+  enabled: true
+  replicas: 1
+  image:
```
kevincai, line 417: can we make this image section omit by default?

helm-charts/charts/kube-starrocks/charts/starrocks/values.yaml
```diff
@@ -408,3 +408,54 @@ configMaps: []
...
+  tolerations: []
+    # - key: "key"
+    #   operator: "Equal|Exists"
+    #   value: "value"
+    #   effect: "NoSchedule|PreferNoSchedule|NoExecute(1.6 only)"
+  # This ConfigMap store the nginx.conf for fe proxy.
+  # If configMapName is empty, operator will use its default nginx.conf.
+  # You can create a ConfigMap with the key nginx.conf to override the default nginx.conf, and please
+  # make sure nginx server is listening on port 8080
+  configMapInfo:
```
kevincai, line 460: any specific configmap we needed to make it running for now?

helm-charts/charts/kube-starrocks/charts/starrocks/values.yaml
yandongxiao: the default value is also kube-dns.kube-system.svc.cluster.local. I will set it to empty string in values.yaml

helm-charts/charts/kube-starrocks/charts/starrocks/values.yaml
yandongxiao: ok, I will check these fields, and make them have default values.

helm-charts/charts/kube-starrocks/charts/starrocks/values.yaml
yandongxiao: operator will create a confimap to store the nginx.conf, mount it to pod

kevincai: we can even generate.the conf on the fly in entrypoint with necessary env var injected by operator.

yandongxiao: > if we are not going to allow user to provide their own nginx conf, is this configmap needed in crd?

no, shall we delete it from values.yaml ?

yandongxiao: operator allow users to use their own nginx.conf, in case our implementation is not comlete.

kevincai: I don't think so, I can't think of any cases that the user needs to modify this nginx.conf. In order not to build our own nginx img, we can do the nginx conf mounted from configmap by the operator. the img is necessary in case they can't access the img from public network.

yandongxiao: ok, I will remove it.

helm-charts/charts/kube-starrocks/values.yaml
```diff
@@ -66,6 +66,7 @@ starrocks:
   initPassword:
     enabled: false
     password: ""
+    # The secret name that contains password, the key of the secret is "password", and you should create it first.
```
kevincai, line 69: should move this into a separate PR.

deploy/operator.yaml
```diff
@@ -23,168 +23,52 @@ rules:
...
   - statefulsets
   verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
+  - '*'
```
kevincai, line 28: is it a good practice to use "*" to acquire all permissions. one of the rule is acquiring only necessary permissions.

helm-charts/charts/kube-starrocks/charts/starrocks/templates/starrockscluster.yaml
```diff
@@ -370,3 +370,33 @@ spec:
       configMapName: {{template  "starrockscluster.cn.configmap.name" . }}
       resolveKey: cn.conf
 {{- end }}
+{{- if .Values.starrocksFeProxySpec.enabled }}
+  starRocksFeProxySpec:
+    {{- if .Values.starrocksFeProxySpec.image.repository }}
```
kevincai, line 387: what if use doesn't provide the repository setting, the crd will not have the image configuration, will operator fill the `image` field?

deploy/operator.yaml
yandongxiao: yes, you are right, but the original permissions are equal '*'.

helm-charts/charts/kube-starrocks/charts/starrocks/templates/starrockscluster.yaml
yandongxiao: in starrocksFeProxySpec, no fields is necessary.

helm-charts/charts/kube-starrocks/values.yaml
yandongxiao: Ok, I will create a new PR for testing for v1.8.0

pkg/sub_controller/feproxy/feproxy_configmap.go
```diff
@@ -0,0 +1,131 @@
...
+		return err
+	}
+	httpPort := rutils.GetPort(config, rutils.HTTP_PORT)
+
+	feServiceName := service.ExternalServiceName(src.Name, src.Spec.StarRocksFeSpec)
+	proxyPass := fmt.Sprintf("http://%s:%d", feServiceName, httpPort)
+
+	resolver := feProxySpec.Resolver
+	if resolver == "" {
+		resolver = "kube-dns.kube-system.svc.cluster.local"
```
kevincai, line 36: whether we can call api server to get the accurate core-dns service? e.g. the core-dns deployed in a different namespace, or the kubernetes has a different domain name other than `.cluster.local`

yandongxiao: I do not think api-server knows that, and that is why I add a resolver filed in CRD

kevincai (approved):

PR #244, opened 2023-08-21
support ports fields

yandongxiao:


helm-charts/charts/kube-starrocks/charts/starrocks/values.yaml
```diff
@@ -63,6 +63,12 @@ starrocksFESpec:
     loadbalancerIP: ""
     # add annotations for fe service.
     annotations: {}
+    # config the service port for fe service.
+    # if you want to use a dedicated port for fe service, you can config the port.
```
kevincai, line 67: we can hava as many links as possible to k8s official document, so if the user doesn't understand the meaning of the configuration, they can learn from k8s official document.

kevincai (approved):

PR #248, opened 2023-08-28
Feature/select appropriate hpa version

yandongxiao:


pkg/apis/starrocks/v1/auto_scale.go
```diff
@@ -54,6 +56,28 @@ const (
 	AutoScalerV2 AutoScalerVersion = "v2"
 )
 
+// Complete completes the default value of AutoScalerVersion
```
kevincai, line 59:
```suggestion
// Complete() completes the default value of AutoScalerVersion
```

pkg/apis/starrocks/v1/auto_scale.go
```diff
@@ -54,6 +56,28 @@ const (
...
+func (version AutoScalerVersion) Complete(major, minor string) AutoScalerVersion {
+	if version != "" {
+		return version
+	}
+	// operator choose a proper default hpa version by checking ths kubernetes version
+	// if kubernetes version > 1.25, use v2 version
```
kevincai, line 65: means >=1.26?

pkg/k8sutils/k8sutils.go
```diff
@@ -287,3 +290,36 @@ func GetConfigMap(ctx context.Context, k8scient client.Client, namespace, name s
...
+
+var (
+	KUBE_MAJOR_VERSION string
+	KUBE_MINOR_VERSION string
+)
+
+// GetKubernetesVersion get kubernetes version. It should not be executed concurrently.
+// The global variable KUBE_MAJOR_VERSION and KUBE_MINOR_VERSION will be set.
+func GetKubernetesVersion() error {
+	config, err := clientcmd.BuildConfigFromFlags("", "~/.kube/config")
```
kevincai, line 302: does this mean that the operator pod needs to be configured with .kube/config?

pkg/k8sutils/k8sutils.go
```diff
@@ -287,3 +290,36 @@ func GetConfigMap(ctx context.Context, k8scient client.Client, namespace, name s
...
+	if err != nil {
+		config, err = rest.InClusterConfig()
+		if err != nil {
+			return err
+		}
+	}
+
+	// 创建一个 discovery.Client 对象，用于查询 API 服务器的元数据
```
kevincai, line 310: make comments english

pkg/apis/starrocks/v1/auto_scale.go
yandongxiao: yes

pkg/k8sutils/k8sutils.go
yandongxiao: It will first check `~/.kube/config`. If it does do not exist, operator will use the the config of starrocks service account.

pkg/apis/starrocks/v1/auto_scale.go
kevincai: let's make it `minorNumber >= 26` which is much intuitive.

pkg/apis/starrocks/v1/auto_scale_test.go
```diff
@@ -0,0 +1,69 @@
...
+			version: AutoScalerV2Beta2,
+			args:    args{},
+			want:    AutoScalerV2Beta2,
+		},
+		{
+			name:    "test for empty version",
+			version: "",
+			args: args{
+				major: "1",
+				minor: "27",
```
kevincai, line 39: shall cover 1.26 as well which is the edge case.

pkg/k8sutils/k8sutils.go
kevincai: shouldn't assume that the .kube/config are never available in a k8s deployment env?

yandongxiao:
By using `clientcmd.BuildConfigFromFlags` ,  We can start or debug operator on localhost.

kevincai (approved):

PR #250, opened 2023-08-29
[Docs] change related fields to release v1.8.1

yandongxiao:


kevincai:
this commit requires an update with new recent PR merges, right?

kevincai (approved):

PR #261, opened 2023-09-14
add namespace flag to make operator watch resources in a single namespace

yandongxiao:

### DIFF
--- a/helm-charts/charts/kube-starrocks/charts/operator/values.yaml
+++ b/helm-charts/charts/kube-starrocks/charts/operator/values.yaml
@@ -48,6 +48,10 @@ starrocksOperator:
     #   operator: "Equal|Exists"
     #   value: "value"
     #   effect: "NoSchedule|PreferNoSchedule|NoExecute(1.6 only)"
-  # if specified, restricts the manager's cache to watch objects in the desired namespace.
+  # If specified, it will restrict operator to watch resources in the specified namespace. And
+  #  1. You must deploy your StarRocks cluster to the specified namespace.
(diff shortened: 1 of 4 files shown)


helm-charts/charts/kube-starrocks/charts/operator/values.yaml
```diff
@@ -47,4 +47,7 @@ starrocksOperator:
     # - key: "key"
     #   operator: "Equal|Exists"
     #   value: "value"
-  #   effect: "NoSchedule|PreferNoSchedule|NoExecute(1.6 only)"
+    #   effect: "NoSchedule|PreferNoSchedule|NoExecute(1.6 only)"
+  # if specified, restricts the manager's cache to watch objects in the desired namespace.
+  # Defaults to all namespaces.
+  watchNamespace: ""
```
kevincai, line 53: what if want to watch multiple namespaces other than all the namespaces?

yandongxiao: k8s support to watch a single namespace or all namespaces. So we can not specify multiple clusters in `watchNamespace` field.

kevincai: that's sound like a risk option exposed to end user, if user have this field set, either set it wrong or the user has multiple clusters deployed in multiple namespace, the operator will not work as desired at all.

Please refine the message here, highlight the risk, if the user doesn't now what it means, don't set it at all.

kevincai (approved):

PR #266, opened 2023-09-19
[Bugfix] support k8s env which has no status.hostIP field

dujl:
In some serverless k8s env ,there is no status.hostIP field, and we do not use hostIp in StarRocks， so remove it.

### DIFF
--- a/pkg/k8sutils/templates/pod/spec.go
+++ b/pkg/k8sutils/templates/pod/spec.go
@@ -227,12 +227,6 @@ func Envs(spec v1.SpecInterface, config map[string]interface{}, feExternalServic
 			},
 		},
 		{
-			Name: "HOST_IP",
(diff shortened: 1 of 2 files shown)


yandongxiao:
Please sign the CLA

yandongxiao:
Direct deletion could lead to backward compatibility issues. Can we determine whether to add this environment variable by inspecting the status.hostIP field?

dujl:
> Direct deletion could lead to backward compatibility issues. Can we determine whether to add this environment variable by inspecting the status.hostIP field?

it is a redundancy field. Env var HOST_IP doesn't be used in starrocks

kevincai:
so the operator cr or helm chart values need an update as well?

yandongxiao (approved):

kevincai (approved):

PR #271, opened 2023-09-22
[Feature] support customize probe

yandongxiao:


pkg/apis/starrocks/v1/load_type.go
```diff
@@ -11,6 +11,9 @@ type loadInterface interface {
 	GetNodeSelector() map[string]string
 	GetAffinity() *corev1.Affinity
 	GetTolerations() []corev1.Toleration
+	GetStartupProbe() *corev1.Probe
```
kevincai, line 14: I would rather allow the user to specify the total tolerant of times for the probe, instead of specifying a different probe, which is only valuable for the users with deep understanding of sr deployment. Let's simplify it.

pkg/apis/starrocks/v1/load_type.go
```diff
@@ -84,6 +85,12 @@ type StarRocksLoadSpec struct {
 	// the reference for configMap which store the config info to start starrocks. e.g. be.conf, fe.conf, cn.conf.
 	// +optional
 	ConfigMapInfo ConfigMapInfo `json:"configMapInfo,omitempty"`
+
+	// StartupProbeFailureSeconds defines the total failure seconds of startupProbe.
+	// By default, the startupProbe use HTTP /api/health Probe, the failureThreshold is 60, the periodSeconds is 5.
```
kevincai, line 90: default failureThreshold is 60 and the periodSeconds is 5, this means the startup will fail if the pod can't start in 300 seconds. Your StartupProbeFailureSeconds is the total time of seconds before startupProbe give up and fail the container start.

pkg/k8sutils/templates/pod/spec.go
```diff
@@ -81,6 +69,19 @@ func getProbe(port int32, path string) corev1.ProbeHandler {
 	}
 }
 
+func completeProbe(failureSeconds *int32, defaultFailureThreshold int32, defaultPeriodSeconds int32,
+	probeHandler corev1.ProbeHandler) *corev1.Probe {
+	probe := &corev1.Probe{}
+	if failureSeconds != nil {
+		probe.FailureThreshold = (*failureSeconds) / defaultPeriodSeconds
```
kevincai, line 76: what if failureSeconds can't be divided by defaultPeriodSeconds, e.g. provided failureSeconds is 1, and defaultPeriodSeconds is 5, should the `FailureThreshold` evaluated to `0` or `1`?

yandongxiao: FailureThreshold will use be the default value, which is 3.

kevincai (approved):

PR #276, opened 2023-10-07
[Docs] add starrocks-fe-and-be.yaml

yandongxiao:

### DIFF
--- a/examples/starrocks/starrocks-fe-and-be.yaml
+++ b/examples/starrocks/starrocks-fe-and-be.yaml
@@ -1,0 +1,18 @@
+apiVersion: starrocks.com/v1
+kind: StarRocksCluster
+metadata:
+  name: starrockscluster-sample
+  namespace: starrocks
+spec:
+  starRocksFeSpec:
+    image: starrocks/fe-ubuntu:latest
+    replicas: 3
+    requests:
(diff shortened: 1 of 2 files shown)


examples/starrocks/starrocks-fe-and-be.yaml
```diff
@@ -0,0 +1,18 @@
...
+  starRocksFeSpec:
+    image: starrocks/fe-ubuntu:latest
+    replicas: 3
+    requests:
+      cpu: 4
+      memory: 16Gi
+  starRocksBeSpec:
+    image: starrocks/be-ubuntu:latest
+    replicas: 3
+    requests:
```
kevincai, line 16: if no limit, what's the uplimit the container will be able to use?

yandongxiao: there is no uplimit for this container, and its [Quality of Service (QoS) classes](https://kubernetes.io/docs/concepts/workloads/pods/pod-qos/) is Burstable

kevincai: so it will likely be OOM if BE process detects the host node mem and use as much of that mem.

kevincai (approved):

PR #284, opened 2023-10-12
[Feature] integration with prometheus

yandongxiao:


helm-charts/charts/kube-starrocks/charts/starrocks/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,47 @@
+{{- if .Values.metrics.serviceMonitor.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
```
kevincai, line 3: 1. is this a k8s builtin resource type? any external dependencies?
2. how does this differs from the service/pod prometheus annotations?

dxyan06: ServiceMonitor is a CRD resource provided by Prometheus. It represents a standard method for metric collection. Compared to the annotation approach, ServiceMonitor allows for more flexible definition of selector and relabeling rules, enabling filtering and modification of metadata for services and endpoints.

yandongxiao: ServiceMonitor is a CRD resource provided by Prometheus. It represents a standard method for metric collection. Compared to the annotation approach, ServiceMonitor allows for more flexible definition of selector and relabeling rules, enabling filtering and modification of metadata for services and endpoints.

yandongxiao: see https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/argocd-server/servicemonitor.yaml

kevincai: what if the user's env is not deployed with prometheus operator and enables this option in our chart values?

yandongxiao: `helm upgrade` will fail.

kevincai: is it a proper behavior?

yandongxiao: In values.yaml, we declared that: 

```
# This configuration is used to integrate with external system Prometheus.
# You can enable the integration by setting the enabled to true.
metrics:
  serviceMonitor:
    # Enable a prometheus ServiceMonitor
    enabled: false
    # Prometheus ServiceMonitor interval
    interval: 15s
```

kevincai: when the helm upgrade/install fails, do you expect the user will know how to fix the failure by installing the prometheus operator?

yandongxiao: we can add more checks, like this:
```
{{- if and (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1") .Values.metrics.serviceMonitor.enabled }}
```

kevincai: ok. update the doc on how to enable prometheus metrics as well since we are redirecting the user to a different approach.

helm-charts/charts/kube-starrocks/charts/starrocks/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,47 @@
...
+metadata:
+  name: {{ template "starrockscluster.name" . }}-fe
+  namespace: {{ template "starrockscluster.namespace" . }}
+  labels:
+    cluster: {{ template "starrockscluster.name" . }}
+spec:
+  endpoints:
+    - port: http
```
kevincai, line 11: does this mean that cr controller will search for the correct port from the given port name of the service?

helm-charts/charts/kube-starrocks/charts/starrocks/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,47 @@
...
+    matchLabels:
+      app.starrocks.ownerreference/name: {{ template "starrockscluster.name" . }}
+      app.kubernetes.io/component: fe
+
+---
+
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ template "starrockscluster.name" . }}-be
```
kevincai, line 29: shall enable cn monitor as well.

helm-charts/charts/kube-starrocks/charts/starrocks/templates/servicemonitor.yaml
yandongxiao: yes

kevincai (approved):

Issue #287, opened 2023-10-17
Add support for running operation on Istio service mesh


vytautas-karpavicius:
Currently there is a problem running Starrocks with operator on Istio service mesh.

After deployment, FE pod starts up, but BE is failing because it can not connect to FE. 
Tthis is related to mysql protocol and mTLS setup on Istio. The issue is documented [here](https://istio.io/latest/docs/ops/deployment/requirements/#server-first-protocols). This can be fixed either by modifying port-name or  adding `appProtocol` as describe in [here](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection).

Currently neither of those options are exposed via operator.

PR #288, opened 2023-10-23
[Enhancement] add appProtocol in fe service

yandongxiao:
Fixes: https://github.com/StarRocks/starrocks-kubernetes-operator/issues/287

### DIFF
--- a/pkg/common/resource_utils/service.go
+++ b/pkg/common/resource_utils/service.go
@@ -33,6 +33,23 @@ const (
 	FeProxyService StarRocksServiceType = "fe-proxy"
 )
 
+const (
+	FeHTTPPortName    = "http"
+	FeRPCPortName     = "rpc"
+	FeQueryPortName   = "query"
+	FeEditLogPortName = "edit-log"
+
+	BePortName          = "be"
+	BeWebserverPortName = "webserver"
+	BeHeartbeatPortName = "heartbeat"
+	BeBrpcPortName      = "brpc"
+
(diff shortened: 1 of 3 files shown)


pkg/common/resource_utils/service.go
```diff
@@ -108,13 +125,20 @@ func BuildExternalService(src *srapi.StarRocksCluster, name string, serviceType
...
-		ports = append(ports, corev1.ServicePort{
+		servicePort := corev1.ServicePort{
 			Name:       sp.Name,
 			Port:       sp.Port,
 			NodePort:   sp.NodePort,
 			Protocol:   corev1.ProtocolTCP,
 			TargetPort: intstr.FromInt(int(sp.ContainerPort)),
-		})
+		}
+		if servicePort.Name == FeQueryPortName {
```
kevincai, line 135: how about the webPort? does it need to add appProtocol `http`?

kevincai: didn't see our webport related to the server-first-protocol document. HTTP protocol doesn't look like a server-first-protocol per the description in the doc.

yandongxiao: MySQL is a server-first-protocol, so we only need to add appProtocol to the query port, 9030.

kevincai (dismissed):

kevincai:
ci operator failure, might need take a look @yandongxiao

kevincai (approved):

Issue #275, opened 2023-10-04
ERROR 1064 (HY000): Access denied for user 'root' (using password: NO) || BE constantly has restart after a few hours


jimenez-f:
I have an issue with Starrocks Helmchart v1.8.3 when I set a password in user root via mysql-client the BE apparently lost the connection via password and constantly has restart the pods of BE with this message:

`
[...]

PR #292, opened 2023-10-27
[Bugfix] remove chart function kube-starrocks.name

yandongxiao:
Fixes: #275 

If starrocks.initPassword.enabled is true, and the value of starrocks.starrocksCluster.name is set, it will cause an issue. When the value starrocks.starrocksCluster.name is set, the service hostname will follows while the hostname inside the initpwd still using the chart name.

### DIFF
--- a/helm-charts/charts/kube-starrocks/charts/operator/templates/_helpers.tpl
+++ b/helm-charts/charts/kube-starrocks/charts/operator/templates/_helpers.tpl
@@ -9,6 +9,6 @@ The default service account name to use for the operator.
 {{- default .Release.Namespace .Values.starrocksOperator.namespaceOverride }}
 {{- end }}
 
-{{- define "kube-starrocks.name" -}}
+{{- define "operator.name" -}}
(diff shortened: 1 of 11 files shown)


kevincai (commented):
what's the consequence of chart upgrade in terms of the name is set or not set?

yandongxiao:
If starrocks.starrocksCluster.name is not set, the result of helm template will remain the same as before. If starrocks.starrocksCluster.name is set and is different from the value of nameOverride, the old configmaps for FE, BE, and CN will be deleted. New configmaps with the new name for FE, BE, and CN will be created. This may result in the restart of FE/BE/CN pods. I need to confirm this again.

yandongxiao:
I have confirmed FE/BE/CN pod will be restarted.

kevincai:
> I have confirmed FE/BE/CN pod will be restarted.

will it affect the statefulset, will the statefulset be deleted and recreated?

yandongxiao:
No. The pod will be restarted because the configmap name of fe.conf will be changed

kevincai (approved):

PR #298, opened 2023-11-02
Feature/multi warehouse 2

yandongxiao:


helm-charts/charts/kube-starrocks/charts/operator/templates/clusterrole.yaml
```diff
@@ -54,18 +54,21 @@ rules:
...
   resources:
   - starrocksclusters
+  - starrockswarehouses
   verbs:
   - '*'
 - apiGroups:
   - starrocks.com
   resources:
   - starrocksclusters/finalizers
+  - starrockswarehouses/finalizers
```
kevincai, line 64: is it still necessary?

yandongxiao: I'm not sure about the purpose of finalizers, so I have kept it consistent. I will confirm later whether this permission is needed.

PR #300, opened 2023-11-03
[Chore] replace label-pr workflow with script

yandongxiao:


scripts/add-function-label-to-pr.sh
```diff
@@ -0,0 +1,44 @@
...
+  fi
+  if [[ $title == *"bugfix"* ]]; then
+    labels+=('bugfix')
+  fi
+  if [[ $title == *"chore"* ]]; then
+    labels+=('chore')
+  fi
+  if [[ $title == *"documentation"* ]]; then
+    labels+=('documentation')
+  fi
```
kevincai, line 34:
```suggestion
  title=$(echo $commit | head -n1 | tr '[:upper:]' '[:lower:]')
  case $title in
    *feature*)
      ;;
    ...
    esac
```

yandongxiao: done

PR #303, opened 2023-11-06
[Bugfix] set proxy_pass_request_body off when redirect to FE

yandongxiao:
This is the original snippet in nginx.conf
```
    location @handle_redirect {
      set $redirect_uri '$upstream_http_location';
      proxy_pass $redirect_uri;
      proxy_set_header Expect $http_expect;
      proxy_pass_request_body on;
      proxy_set_header Host $host;
      proxy_set_header X-Real-IP $remote_addr;
      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
      error_page 307 = @handle_redirect;
    }
```

When Nginx forwards requests to Leader FE and then BE, it follows the above  Nginx configuration. Because proxy_pass_request_body is set to true, Nginx sends requests to Leader FE. When FE returns 307, Nginx forwards the request to BE. At this point, the data cached in Nginx is lost, causing the error.

```
2023/11/03 14:35:39 [error] 23#23: *23 sendfile() failed (32: Broken pipe) while sending request to upstream, client: 10.244.0.26, server: , request: "PUT /api/transaction/load HTTP/1.1", upstream: "http://10.96.43.63:8030/api/transaction/load", host: "10.244.0.41:8080"
```

### DIFF
--- a/pkg/sub_controller/feproxy/feproxy_configmap.go
+++ b/pkg/sub_controller/feproxy/feproxy_configmap.go
@@ -28,8 +28,9 @@ func (controller *FeProxyController) SyncConfigMap(ctx context.Context, src *sra
 	}
 	httpPort := rutils.GetPort(config, rutils.HTTP_PORT)
 
-	feServiceName := service.ExternalServiceName(src.Name, src.Spec.StarRocksFeSpec)
-	proxyPass := fmt.Sprintf("http://%s:%d", feServiceName, httpPort)
+	feSearchServiceName := service.SearchServiceName(src.Name, feSpec)
+	feExternalServiceName := service.ExternalServiceName(src.Name, feSpec)
+	proxyPass := fmt.Sprintf("http://%s:%d", feExternalServiceName, httpPort)
 
 	resolver := feProxySpec.Resolver
 	if resolver == "" {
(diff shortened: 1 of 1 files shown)


pkg/sub_controller/feproxy/feproxy_configmap.go
```diff
@@ -113,6 +114,26 @@ http {
     }
 
     location @handle_redirect {
+      if ($upstream_http_location ~ "%[3]s") {
```
kevincai, line 117: how about
```
set $pass_body on;
if ( xxx ) {
    set $pass_body off;
}

proxy_pass_request_body $pass_body;
```

yandongxiao: `if` can be used in rewrite context. See https://nginx.org/en/docs/http/ngx_http_rewrite_module.html

kevincai: just need to check if it is a FE redirect and set $pass_body accordingly, no need to do rewrite.

yandongxiao: yes, I tried. `if` is not allowed to use without rewrite.

kevincai (approved):

xlfjcg:
client should not send body before accept 100, The only exception should be response timeouts

yandongxiao:
> client should not send body before accept 100, The only exception should be response timeouts

I think what you mean is: nginx should not send request body before it accept 307 to BE ?

PR #312, opened 2023-11-20
[Chore] change related fields to release v1.8.7

yandongxiao:


helm-charts/charts/kube-starrocks/values.yaml
```diff
@@ -255,8 +255,22 @@ starrocks:
...
+  
+    # StartupProbeFailureSeconds defines the total failure seconds of startup Probe.
+    # default value is 300 seconds
+    startupProbeFailureSeconds:
+    # LivenessProbeFailureSeconds defines the total failure seconds of liveness Probe.
+    # default value is 15 seconds
+    livenessProbeFailureSeconds:
+    # ReadinessProbeFailureSeconds defines the total failure seconds of readiness Probe.
+    # default value is 15 seconds
+    readinessProbeFailureSeconds:
```
kevincai, line 272: is it OK to leave the value empty? looks like weird for an integer type.

yandongxiao: This is done intentionally. If we set a default value, users with versions <=v1.8.6 would encounter upgrade failures during their next deployment.

kevincai (approved):

PR #317, opened 2023-11-21
[Enhancement] update warehouse status

yandongxiao:
export args for operator helm chart and update warehouse status

### DIFF
--- a/helm-charts/charts/warehouse/templates/configmap.yaml
+++ b/helm-charts/charts/warehouse/templates/configmap.yaml
@@ -6,7 +6,7 @@ metadata:
   namespace: {{ template "starrockswarehouse.namespace" . }}
   labels:
     warehouse: {{ template "starrockswarehouse.name" . }}
-    app: "cn"
+    app: "warehouse"
 data:
 {{ include "starrockswarehouse.config" . | indent 2 }}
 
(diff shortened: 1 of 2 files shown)


helm-charts/charts/kube-starrocks/charts/operator/values.yaml
```diff
@@ -33,6 +33,8 @@ starrocksOperator:
     # image sliced by "repository:tag"
     repository: starrocks/operator
     tag: v1.8.6
+  args:
```
kevincai, line 36: not necessary?

yandongxiao: Operator has some args, like -v for debug log, --namespace

kevincai: I think the namespace has already exposed?. debug log is just for troubleshoot? the example --elect-leader is confusing.

yandongxiao: We can hide the --elect-leader in template.

yandongxiao: I have remove the args from values.yaml

kevincai (approved):

PR #324, opened 2023-11-27
[Enhancement] Support to mount emptyDir in storageVolumes

yandongxiao:


helm-charts/charts/kube-starrocks/charts/starrocks/values.yaml
```diff
@@ -154,6 +154,14 @@ starrocksFESpec:
     storageSize: 10Gi
     # Setting this parameter can persist log storage
     logStorageSize: 5Gi
+  # mount other volumes if necessary.
+  # Note: please use storageSpec field for persistent metadata and log.
+  storageVolumes: []
```
kevincai, line 159: the user will be confused by this `storageVolumes` and `storageSpec`, what's the different when to use which one.

yandongxiao: yes, do you have any suggestions? 
storageVolumes can be configured with PVC and EmptyDir, or should we name it emptyDirVolume and only allow users to set EmptyDir?

helm-charts/charts/kube-starrocks/charts/starrocks/values.yaml
```diff
@@ -595,3 +613,8 @@ starrocksFeProxySpec:
   # ReadinessProbeFailureSeconds defines the total failure seconds of readiness Probe.
   # default value is 15 seconds
   readinessProbeFailureSeconds:
+  # Note: will create emptyDir volume for fe proxy, PVC is not supported.
+  emptyDirs: []
```
kevincai, line 617: what if the `emptyDirs` has a path that conflicts with the storageSpec path?

kevincai: that's what I am afraid of. statefulset refuses the change but the error will be silent to the user, the user is waiting for something happen but never happens, check if we can validate this kind of errors in operator?

yandongxiao: OK, a good suggestion, will check it. And we can also show errors(or k8s event) from statefulset and pod  on StarRocksCluster CR. So we can describe the StarRocksCluster CR to get the error.

kevincai: right! When the mount path conflicts from multiple configurations, what we can do for now is to refuse the crd apply, or show errors in cr status at least.

yandongxiao: Operator can not refuse the crd apply, we should use k8s webhook to achive this.

kevincai (approved):

PR #341, opened 2023-12-06
[Enhancement] use zap as logger

### LINKED ISSUE #342: [Enhancement] use zap to replcace klog package
## Describe the current behavior

The current Operator directly uses klog as the logging library. The main issues are:
1. klog does not support preset fields. Each klog.Info log contains redundant fields such as namespace, name, etc., and some fields may even be omitted. The readability of the final log entries is not high.
2. klog does not support the JSON format, which is not friendly for log collection.
3. The controller-runtime that the Operator depends on uses the zap log format by default, resulting in two log formats in the logs.

## Describe the enhancement

1. **Provide Debug function.**
2. **Support preset log fields.**
    1. There is no need to specify these contents repeatedly in each print statement. For example, the Name, Namespace, etc., of the StarRocksCluster CRD.
    2. At any position in the code, you can dynamically decide to add more preset field information. Preset fields include two types: Which Controller is performing which reconciliation actions.
        1. Controllers include FE, BE, CN, FeProxy.
        2. Actions can refer to the actions in the WarehouseSubController and ClusterSubController interfaces.
3. **Log content.**
    1. In addition to the context of the log print, such as namespace, name, controller, action, etc., it also includes: Log level, timestamp, logger, caller, message.
4. **Where to print logs.**
    1. Print logs when the Sub Controller starts to reconcile.
    2. Where to print logs? Print logs when calling the k8s API.
5. **Log format: Support text and JSON formats, default to text format.** 
    1. Text format: ./main --zap-encoder console
    2. JSON format: ./main --zap-encoder json

Controller-runtime has its own log package under its pkg/log, and it also has an implementation pkg/log/zap. The zap package is implemented based on go.uber.org/zap. In general, we can use this log package too.

yandongxiao:
close #342


kevincai (approved):

deploy/operator.yaml
```diff
@@ -182,6 +182,8 @@ spec:
         - /sroperator
         args:
         - --leader-elect
+        - --zap-time-encoding iso8601
+        - --zap-encoder console
```
matheusjerico, line 186: This parameters are broken.. If I try to install the operator using this `deploy/operator.yaml` I am receiving this error:
```bash
kubectl logs kube-starrocks-operator-b49c6d4fc-sljzj -f -p 
flag provided but not defined: -zap-time-encoding iso8601
Usage of /sroperator:
```

To have success to install the operator, I need to remove the both lines:
```bash
        - --zap-time-encoding iso8601
        - --zap-encoder
```

yandongxiao: Thanks, I will try to fix it.

Issue #384, opened 2024-01-03
Operator should skip syncing the fields that is not defined in the CRD


dengliu:
## Describe the bug

Currently operator consider all the fields that are not defined in CRD as `null` and try to sync the `null` value to the object.
In case where the object were created via other external controller e.g. load balancer, it will cause failure.
E.g. a fe-service loadbalancer that were provisioned by [aws load balancer controller](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.6/guide/service/nlb/) will have ` loadBalancerClass: service.k8s.aws/nlb` field added by the controller. 
while the operator considers it's has `null` value and try to sync the value and fails:

```
FeController Sync create or patch external service namespace celostar-dic name celostar-fe-service failed, message Service "celostar-fe-service" is invalid: spec.loadBalancerClass: Invalid value: "null": may not change once set

StarRocksClusterReconciler reconcile sub resource reconcile failed, namespace=celostar-dic, name=celostar, controller=feController, error=Service "celostar-fe-service" is invalid: spec.loadBalancerClass: Invalid value: "null": may not change once set
```
## To Reproduce

1) Install  [aws load balancer controller](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.6/guide/service/nlb/) to the k8s cluster
2) Annotate fe-service with these annotation:
```
starrocksFESpec set to
service:
  type: LoadBalancer
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-name: celostar
    service.beta.kubernetes.io/aws-load-balancer-type: external
    service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
    service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: instance
    service.beta.kubernetes.io/aws-load-balancer-target-group-attributes: stickiness.enabled=true,stickiness.type=source_ip

```

## Expected behavior
Operator should skip syncing the fields that is not defined in the CRD
## Please complete the following information

- Operator Version: [e.g. v1.8.6]
- Chart Name: [e.g. kube-starrocks]
- Chart Version [e.g. v1.8.6]

dengliu:
@yandongxiao 
One of the idea is to maintain a hash table of the fields that defined in CRD spec then skip syncing the field that is not in the hash table.

PR #387, opened 2024-01-05
[BugFix] use Patch not Update to modify service object

yandongxiao:
Fixes: #384 

This is an update strategy based on the JSON Merge Patch standard. JSON Merge Patch merges the submitted JSON object into the existing resource object. If a field exists in the submitted JSON object, then the value of this field in the resource object will be replaced. If a field does not exist in the submitted JSON object, then the value of this field in the resource object will remain unchanged.


dengliu (approved):

dengliu:
@yandongxiao when will be a new version of helm chart release that contains this fix? Thanks!

yandongxiao:
In the next one to two weeks.

PR #394, opened 2024-01-11
[Enhancement] disable FE scale to 1

yandongxiao:
If scale to 1, Operator will report an error. 

```
k get starrocksclusters.starrocks.com kube-starrocks
NAME             PHASE    FESTATUS   BESTATUS   CNSTATUS   FEPROXYSTATUS
kube-starrocks   failed   running               running

k get starrocksclusters.starrocks.com kube-starrocks -oyaml | grep reason
reason: 'error from FE controller: the replicas of statefulset kube-starrocks-fe can not be scaled to 1'
```


yandongxiao:
StarRocks uses BDBJE to select leader for FE, and I did not test this case. 
I think you'd better to scale FE one by one.

Issue #424, opened 2024-01-28
Add a global `image:tag` field in SR Helm chart


mingmxu:
## Describe the current behavior

Currently every sub-component(`starrocksFESpec`, `starrocksCnSpec`, `starrocksBeSpec`, `starrocksFeProxySpec`) has its own `image:tag` field.  

## Describe the enhancement

Add a new field `starrocksCluster.imageTag` as the default value, the existing one could be kept in case there's a need to override it.

## Additional context

Practically, in one cluster it should always use the same image_tag, built from the same snapshot of code base. It's redundant to set them by sub-component, error-prone to miss one line.

cc @dengliu

yandongxiao:
In the individual spec definitions of the sub-components, there are indeed many fields with the same name, such as image, imagePullSecrets, nodeSelector, schedulerName. Generally, the contents of these fields are consistent in the individual spec definitions of the sub-components. Of course, it would be better if a way could be found to set default values uniformly.

yandongxiao:
work on it.

dengliu:
`nodeSelector` can be different as FE and BE usually require different node spec.

PR #425, opened 2024-01-29
[Enhancement] Define some default values uniformly

yandongxiao:
# Description

In the individual spec definitions of the sub-components, there are indeed many fields with the same name, such as image, imagePullSecrets, nodeSelector, schedulerName. Generally, the contents of these fields are consistent in the individual spec definitions of the sub-components. 

Fixes: #424 

# Checklist

  file of starrocks chart.
  chart( kube-starrocks chart).

### DIFF
--- a/helm-charts/charts/kube-starrocks/charts/starrocks/templates/_helpers.tpl
+++ b/helm-charts/charts/kube-starrocks/charts/starrocks/templates/_helpers.tpl
@@ -240,3 +240,332 @@ the first 8 digits are taken, which will be used as the annotations for pods.
 {{- end }}
 {{- end }}
 
+{{/*
+Get the value of the schedulerName field in the starrocksFESpec
+*/}}
+{{- define "starrockscluster.fe.schedulerName" -}}
+{{- if .Values.starrocksFESpec.schedulerName -}}
+{{- .Values.starrocksFESpec.schedulerName -}}
+{{- else if .Values.starrocksCluster.componentValues.schedulerName -}}
+{{- .Values.starrocksCluster.componentValues.schedulerName -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Get the value of the schedulerName field in the starrocksBeSpec
+*/}}
+{{- define "starrockscluster.be.schedulerName" -}}
+{{- if .Values.starrocksBeSpec.schedulerName -}}
+{{- .Values.starrocksBeSpec.schedulerName -}}
+{{- else if .Values.starrocksCluster.componentValues.schedulerName -}}
+{{- .Values.starrocksCluster.componentValues.schedulerName -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Get the value of the schedulerName field in the starrocksCnSpec
+*/}}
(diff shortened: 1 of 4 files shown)


helm-charts/charts/kube-starrocks/charts/starrocks/values.yaml
```diff
@@ -61,7 +106,7 @@ starrocksFESpec:
   image:
     # image sliced by "repository:tag"
     repository: starrocks/fe-ubuntu
-    tag: 3.1-latest
+    tag: ""
```
dengliu, line 109: if we set it to empty string. will it also take precedence over the values in `componentValues`?

dengliu (commented):
+ @mingmxu to take a loot as well

helm-charts/charts/kube-starrocks/charts/starrocks/values.yaml
yandongxiao: No. Only when it has a non-empty value.

dengliu (approved):
Thanks for the change. This is super useful!

helm-charts/charts/kube-starrocks/charts/starrocks/values.yaml
```diff
@@ -53,6 +53,51 @@ starrocksCluster:
...
+    # affinity for pod scheduling.
+    affinity: {}
+      # nodeAffinity:
+      #   requiredDuringSchedulingIgnoredDuringExecution:
+      #     nodeSelectorTerms:
+      #     - matchFields:
+      #       - key: metadata.name
+      #         operator: In
+      #         values:
+      #         - target-host-name
```
mingmxu, line 100: Let's add `podLabels` as well.

yandongxiao: commited

helm-charts/charts/kube-starrocks/charts/starrocks/templates/_helpers.tpl
```diff
@@ -240,3 +240,332 @@ the first 8 digits are taken, which will be used as the annotations for pods.
...
+{{- else if .Values.starrocksCluster.componentValues.tolerations -}}
+{{- toYaml .Values.starrocksCluster.componentValues.tolerations -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Get the value of the nodeSelector field in the starrocksFESpec
+*/}}
+{{- define "starrockscluster.fe.nodeSelector" -}}
+{{- if .Values.starrocksFESpec.nodeSelector -}}
```
kevincai, line 379: this is a complete replace behavior, not the merge together behavior, need explicit this in the values.yaml comments.

Otherwise, the user can define some of the common criteria in compontValues and add additional specific criteria in each XXSpec and expect the final result is the merge-replace output.

The same applies to hostAlias/Labels/affinity ...

yandongxiao: committed

kevincai (approved):

Issue #448, opened 2024-02-15
[enhancement] CRD support in sr-operator Helm chart


mingmxu:
## Describe the current behavior

StarRocksCluster CRD is saved in a separated folder as shown in [here](https://github.com/StarRocks/starrocks-kubernetes-operator/tree/main/helm-charts/charts/kube-starrocks/charts/operator).  
In this structure, there's no upgrade support on CRD, since Helm by default ignore the CRD folder simply when the CRD is there already. As a result it can be handled in a delete-recreate pattern only. Implicitly there's always downtime of SR clusters since they are deleted when the old CRD is deleted, plus manually re-deploy the SR clusters after the new CRD is created.

## Describe the enhancement

Moving CRD into the regular [templates](https://github.com/StarRocks/starrocks-kubernetes-operator/tree/main/helm-charts/charts/kube-starrocks/charts/operator/templates) folder, or moving it to another sub-chard as dependency, CRD upgrade is treated as other resources.

## Additional context

SR operator/CRD is the mainstream pattern which has a global CRD and a single global controller for the CRD. It's the `perfectly legitimate way of packaging` as described [here](https://github.com/helm/community/blob/main/hips/hip-0011.md#no-special-crd-management).

Ps, CRD versioning is another topic in https://github.com/StarRocks/starrocks-kubernetes-operator/issues/445. 

cc @dengliu @yandongxiao @kevincai

yandongxiao:
I think you are right. we can move the CRD yaml to templates directory.

yandongxiao:
Currently, the Operator manages two CRDs, namely [starrocksclusters](https://github.com/StarRocks/starrocks-kubernetes-operator/blob/main/helm-charts/charts/kube-starrocks/charts/operator/crds/starrocks.com_starrocksclusters.yaml) and [starrockswarehouses](https://github.com/StarRocks/starrocks-kubernetes-operator/blob/main/helm-charts/charts/warehouse/crds/starrocks.com_starrockswarehouses.yaml). These two CRDs are placed in the `crds` directory of the Helm Chart. Due to the incomplete definition of CRD by Operator, we always receive requests from users to add support for certain fields, which often leads to changes in CRD.

[...]

PR #449, opened 2024-02-19
[WIP][Enhancement] move CRD yaml file to templates directory

yandongxiao:
# Related Issue(s)
fixes: #448


mingmxu (approved):

yandongxiao:
There are errors when execute `helm upgrade`.

Error: UPGRADE FAILED: rendered manifests contain a resource that already exists. Unable to continue with update: CustomResourceDefinition "starrocksclusters.starrocks.com" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "starrocks"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "starrocks"

mingmxu:
I'm surprised that helm doesn't add the annotations when it's saved in `/crds` folder. As `helm upgrade` is not a valid way so far, I won't call it a break change, but it does worth to highlight in the next release version that:
* from v1.9.3 and later, `helm upgrade` is supported;
* for v1.9.2 or earlier -> v1.9.3, needs to run the two commands manually.

yandongxiao:
> I'm surprised that helm doesn't add the annotations when it's saved in `/crds` folder. As `helm upgrade` is not a valid way so far, I won't call it a break change, but it does worth to highlight in the next release version that:
> 
> * from v1.9.3 and later, `helm upgrade` is supported;
> * for v1.9.2 or earlier -> v1.9.3, needs to run the two commands manually.

I think your suggestion is acceptable. @kevincai

yandongxiao:
‌‌Close this PR first to prevent it from being merged.

PR #450, opened 2024-02-21
[Enhancement] support be_http_port and be_port in conf

yandongxiao:
# Description

In the latest StarRocks version, the port names for BE and CN have been modified. The good news is that the old port names can still be used. However, if users use the new port names, Operators may not recognize them, leading to Operators potentially using default ports as liveness and readiness ports. This could result in container startup failures.

# Checklist

  file of starrocks chart.
  chart( kube-starrocks chart).

### DIFF
--- a/pkg/common/resource_utils/configmap.go
+++ b/pkg/common/resource_utils/configmap.go
@@ -34,9 +34,18 @@ const (
 
 // the cn or be ports key
 const (
-	THRIFT_PORT            = "thrift_port"
-	BE_PORT                = "be_port"
-	WEBSERVER_PORT         = "webserver_port"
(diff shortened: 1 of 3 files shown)


pkg/common/resource_utils/configmap.go
```diff
@@ -84,6 +94,15 @@ func GetPort(config map[string]interface{}, key string) int32 {
 		if port, err := strconv.ParseInt(v.(string), 10, 32); err == nil {
 			return int32(port)
 		}
+	} else if key == THRIFT_PORT {
```
kevincai, line 97: should ignore the thrift_port config if thrift_port == 0?

kevincai (approved):

PR #455, opened 2024-02-23
[BugFix] Ensure query port can be configured in init-password job

yandongxiao:
# Description

A scenario that requires modifying the container listening query port:

1. The source cluster is deployed on virtual machines, while the target cluster is deployed on k8s. External access to the internal IP of k8s is not possible, but pods within k8s can access external IPs.
2. The migration tool is deployed outside of k8s. The StarRocks cluster on k8s is exposed through /etc/host (frontend domain name, node IP of k8s) + NodePort.
3. In the cross-cluster migration tool, use the port obtained from `Show Frontend` for recording in Frontend. At the same time, use this port to access the target cluster (instead of configuring target_fe_query_port).
4. The target cluster pulls data from the source cluster.

# Checklist

  file of starrocks chart.
  chart( kube-starrocks chart).

### DIFF
--- a/helm-charts/charts/kube-starrocks/charts/starrocks/templates/_helpers.tpl
+++ b/helm-charts/charts/kube-starrocks/charts/starrocks/templates/_helpers.tpl
@@ -213,6 +213,20 @@ the first 8 digits are taken, which will be used as the annotations for pods.
   {{- end }}
 {{- end }}
 
+{{- define "starrockscluster.fe.query.port" -}}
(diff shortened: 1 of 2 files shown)


yandongxiao:
Tested three scenarios:
1. Configuration does not include the query port setting.
2. 2. Configuration includes the query port setting.
3. The query port setting in the configuration is commented out.
All three scenarios meet expectations.

PR #486, opened 2024-03-21
[Documentation] Update mount_persistent_volume_howto.md

yandongxiao:


doc/mount_persistent_volume_howto.md
```diff
@@ -124,3 +121,23 @@ starrocks:
 
 See [Install StarRocks by kube-starrocks chart](../helm-charts/charts/kube-starrocks/README.md) to learn how to deploy
 StarRocks Operator and Cluster
+
+## 3. Some Special storageClassName
+
+Normally, the `storageClassName` is the name of the StorageClass that you want to use for the PersistentVolumeClaim.
+We have also provided some special `storageClassName` for you to use:
+
+1. `emptyDir`. It is a good choice when you want mount a volume to a special directory in the container, e.g. /tmp.
```
kevincai, line 130:
```suggestion
1. `emptyDir`. It is a good choice when you want to mount a volume into the container for temporary usage, e.g. /tmp. Be aware that the files and directories written to the volume will be completely lost upon container restarting.
```

doc/mount_persistent_volume_howto.md
```diff
@@ -124,3 +121,23 @@ starrocks:
...
+2. `hostPath`. It is a good choice when you want to use local storage on the node. And you should specify the `hostPath`
```
kevincai, line 131:
```suggestion
2. `hostPath`. It is a good choice when you want to the host's storage for the container, the data will be still there as along as the container is still running on the host. The data will be unavailable upon the container rescheduling to a different host. The typical scenario is to use it as cache volume. The `hostPath` field is required when this type is used.
```

kevincai (approved):

PR #499, opened 2024-04-01
[Feature] Support init Containers

yandongxiao:
# Description

In some cases, we need to update the configuration of the host machine. For example, in order to collect Core Dump files, we need to set `/proc/sys/kernel/core_pattern` in the host environment. There are two options available:
1. Start StarRocks Component with privileged root permission and run it continuously with that permission. This requires the Operator to have the ability to change the entrypoint script.
2. Before starting StarRocks Component, start a privileged root init container first. This method provides higher security.

# Checklist


kevincai (approved):

yandongxiao:
‌‌Should be by the end of this week or early next week.

PR #546, opened 2024-06-14
feat: add support for TopologySpreadConstraints

davidspek:
# Description

This PR adds support for using Pod Topology Spread Constraints. This is useful when trying to distribute the scheduling of pods evenly across failure domains such as nodes or availability zones. In many situations this cannot be done with simple Affinity controls.

# Related Issue(s)

Please list any related issues and link them here.

# Checklist

  file of starrocks chart.
  chart( kube-starrocks chart).

### DIFF
--- a/pkg/k8sutils/templates/pod/spec.go
+++ b/pkg/k8sutils/templates/pod/spec.go
@@ -265,6 +265,7 @@ func Spec(spec v1.SpecInterface, container corev1.Container, volumes []corev1.Vo
 		ServiceAccountName:            spec.GetServiceAccount(),
 		TerminationGracePeriodSeconds: spec.GetTerminationGracePeriodSeconds(),
 		Affinity:                      spec.GetAffinity(),
+		TopologySpreadConstraints:     spec.GetTopologySpreadConstraints(),
 		Tolerations:                   spec.GetTolerations(),
 		ImagePullSecrets:              spec.GetImagePullSecrets(),
 		NodeSelector:                  spec.GetNodeSelector(),
--- a/config/crd/bases/starrocks.com_starrocksclusters.yaml
+++ b/config/crd/bases/starrocks.com_starrocksclusters.yaml
@@ -4172,6 +4172,184 @@ spec:
                           type: string
                       type: object
                     type: array
+                  topologySpreadConstraints:
+                    description: |-
+                      (Optional) TopologySpreadConstraints for spreading pods across failure-domains
+                      More info: https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/
+                    items:
+                      description: TopologySpreadConstraint specifies how to spread
+                        matching pods among the given topology.
+                      properties:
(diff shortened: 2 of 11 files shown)


pkg/k8sutils/templates/pod/spec.go
```diff
@@ -265,6 +265,7 @@ func Spec(spec v1.SpecInterface, container corev1.Container, volumes []corev1.Vo
 		ServiceAccountName:            spec.GetServiceAccount(),
 		TerminationGracePeriodSeconds: spec.GetTerminationGracePeriodSeconds(),
 		Affinity:                      spec.GetAffinity(),
+		TopologySpreadConstraints:     spec.GetTopologySpreadConstraints(),
```
yandongxiao, line 268: Since this is a newer feature on k8s. If the user is using an older version of k8s, such as version 1.18, will it cause deployment to fail?

yandongxiao (commented):
Thank you for your PR!

pkg/k8sutils/templates/pod/spec.go
yandongxiao: I have verified it.
```
# deployments.apps "kube-starrocks-operator" was not valid:
# * patch: Invalid value: "map[spec:map[aaaa:[]]]": strict decoding error: unknown field "spec.aaaa"```

yandongxiao:
‌‌Could you share the scenarios in which you use TopologySpreadConstraints?

davidspek:
You need TopologySpreadConstraints if you want to evenly distribute pods across availability zones in scenarios where you might have more pods than the amount of availability zones. This is especially important since in most cloud providers the persistent disks are bound to availability zones. Meaning that if on the first time the pods get scheduled they aren't spread across all the availability zones you will not be able to correct that later (without serious manual effort). A pod anti-affinity rule can work, but only works if the amount of zones is equal to the amount of pods you will ever want to schedule.

pkg/k8sutils/templates/pod/spec.go
davidspek: TopologySpreadContraints was promoted to beta in Kubernetes 1.18 as far as I can find so quickly. Which was released 4 years ago, or almost half the life of Kubernetes :). I think most projects have given up on supporting such old versions of Kubernetes since it can prevent projects from moving forwards with features. It might be possible to have this feature only get enabled after doing a Kubernetes version check, but I'm not familiar with the code enough to do that quickly, nor am I sure it's worth the effort for maintaining support for Kubernetes versions that people shouldn't have been using anymore for years now.

yandongxiao: https://kubernetes.io/blog/2020/05/introducing-podtopologyspread/
From this blog, topologySpreadConstraints was introduced in 1.18, which is the oldest version starrocks operator support.

yandongxiao (approved):

Hatlassian:
This is awesome, thanks for this!

Issue #485, opened 2024-03-20
Can't change an existing cluster to `hostPath`


nferrario:
## Describe the bug

Changing the storage configuration for an existing cluster gives the following error:

```
2024-03-20T14:22:28.787Z    ERROR    StarRocksClusterReconciler    sub controller reconciles spec failed    {"name": "starrocks", "namespace": "starrocks", "subController": "cnController", "error": "StatefulSet.apps \"starrocks-cn\" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden"}
github.com/StarRocks/starrocks-kubernetes-operator/pkg/controllers.(*StarRocksClusterReconciler).Reconcile
    /go/src/app/pkg/controllers/starrockscluster_controller.go:94
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
    /go/src/app/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:121
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
    /go/src/app/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:320
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
    /go/src/app/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:273
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
    /go/src/app/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:234
[...]

PR #552, opened 2024-06-24
[Enhancement] If hostPath is set, the storage class will be considered as hostPath

yandongxiao:
# Description

# Related Issue(s)

Fixes: #485 

# Checklist

### DIFF
--- a/pkg/k8sutils/templates/pod/mount.go
+++ b/pkg/k8sutils/templates/pod/mount.go
@@ -13,10 +13,6 @@ import (
 // SpecialStorageClassName returns the special storage class name of the storage volume, else return "".
 // Now we support HostPath and EmptyDir as special storage class.
 func SpecialStorageClassName(sv v1.StorageVolume) string {
-	if sv.HostPath != nil {
-		return v1.HostPath
-	}
-
 	storageClassName := sv.StorageClassName
 	if storageClassName != nil {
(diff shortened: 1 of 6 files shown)


pkg/k8sutils/templates/pod/mount.go
```diff
@@ -10,9 +10,23 @@ import (
 	"github.com/StarRocks/starrocks-kubernetes-operator/pkg/common/hash"
 )
 
-func IsSpecialStorageClass(storageClassName *string) bool {
-	return storageClassName != nil && (common.EqualsIgnoreCase(*storageClassName, v1.EmptyDir) ||
-		common.EqualsIgnoreCase(*storageClassName, v1.HostPath))
+// SpecialStorageClassName returns the special storage class name of the storage volume, else return "".
+// Now we support HostPath and EmptyDir as special storage class.
+func SpecialStorageClassName(sv v1.StorageVolume) string {
+	if sv.HostPath != nil {
```
kevincai, line 16: what if the the sv.HostPath is provided and the sv.StorageClassName is provided as well and they mismatch?

yandongxiao: Good catch. I think the storage class name will have a high priority.

kevincai (approved):

PR #576, opened 2024-08-07
feat: add resources option for init_job

anthony974:
# Description

Add in helm chart possibility to add resources configuration for the init_password pod.

We are using openshift, on our cluster we have some kyverno rules set and the need-containers-ressources rules is applied.

We wanted to change the default password but the init_job container coudln't be deployed because the limits/requests wasn't set.

# Checklist

  file of starrocks chart.
  chart( kube-starrocks chart).


bastiansmn (approved):
LGTM

buu-nguyen:
This changes is straight forward. LGTM.
Please take a look at it, thanks Team.

buu-nguyen (approved):

yandongxiao (approved):

Issue #494, opened 2024-04-01
[helm]support config override in starrocksFESpec/starrocksBeSpec/starrocksCnSpec


mingmxu:
## Describe the current behavior

Currently all the `config` in starrocks helm chart are taken as a block, see the code [here](https://github.com/StarRocks/starrocks-kubernetes-operator/blob/main/helm-charts/charts/kube-starrocks/charts/starrocks/templates/_helpers.tpl#L53-L72). As a result, partial update is not supported.

## Describe the enhancement

Support partial update in `values.yaml` to simplify configuration changes. 

## Additional context

cc @dengliu @yandongxiao

yandongxiao:
‌‌The "partial update" you mentioned refers to adding or updating some additional configurations in other values.yaml files on top of a certain base values.yaml configuration?

mingmxu:
> ‌‌The "partial update" you mentioned refers to adding or updating some additional configurations in other values.yaml files on top of a certain base values.yaml configuration?

Yes,

yandongxiao:
This is a bit challenging because the current type of config is a string. We need to parse this string in _helper, convert it into the format of a map, and finally merge it with the default values. I tried to write it, but failed.

PR #580, opened 2024-08-20
[helm]support config override in starrocksFESpec/starrocksBeSpec/starrocksCnSpec

dengliu:
# Description
https://github.com/StarRocks/starrocks-kubernetes-operator/issues/494

# Test
### 1) Test config spec setting
with values1.yaml, run 
```
helm template --debug kube-starrocks --values kube-starrocks/values.yaml --values kube-starrocks/values-1.yaml

```
```
# spec for component be, provide storage and compute function.
starrocks:
  starrocksFESpec:
    configyaml:
      query_port: 9031
      foo: FEAdditionalValue
      additonal: values
  starrocksBeSpec:
    configyaml:
      be_port: 9061
      foo: BEAdditionalValue
  starrocksCluster:
    enabledCn: true
  starrocksCnSpec:
    configyaml:
      heartbeat_service_port: 9051
      foo: CNAdditionalValue
```
Got 
image
image
image

### 2) Test config hash
with template below, a minor change in the configyaml values generate totally different hash value
```
fe-config-hash: "{{template "starrockscluster.fe.config.hash" . }}"
be-config-hash: "{{template "starrockscluster.be.config.hash" . }}"
cn-config-hash: "{{template "starrockscluster.cn.config.hash" . }}"

```
image
image
image

### 3) Test with quote in values
```
  starrocksFESpec:
    configyaml:
      LOG_DIR : ${STARROCKS_HOME}/log
      DATE : '"$(date +%Y%m%d-%H%M%S)"'
      JAVA_OPTS: '"-Dlog4j2.formatMsgNoLookups=true -Xmx8192m -XX:+UseG1GC -Xlog:gc*:${LOG_DIR}/fe.gc.log.$DATE:time"'
      mysql_service_nio_enabled: false
```
image
### 4) Test with configyaml not setting
Verified that it uses the string value in config field

### 5) Verified default value setting from configyaml
image
image
image

# Related Issue(s)

N/A

# Checklist

  file of starrocks chart.
  chart( kube-starrocks chart).

### DIFF
--- a/helm-charts/charts/kube-starrocks/charts/starrocks/templates/_helpers.tpl
+++ b/helm-charts/charts/kube-starrocks/charts/starrocks/templates/_helpers.tpl
@@ -52,22 +52,40 @@ starrockscluster
 
 {{- define "starrockscluster.fe.config" -}}
 fe.conf: |
-{{- if .Values.starrocksFESpec.config }}
-{{- .Values.starrocksFESpec.config | nindent 2 }}
+{{- if and .Values.starrocksFESpec.configyaml (kindIs "map" .Values.starrocksFESpec.configyaml) }}
+  {{- range $key, $value := .Values.starrocksFESpec.configyaml }}
+    {{ $key }} = {{ $value }}
+  {{- end }}
+{{- else if .Values.starrocksFESpec.configyaml }}
+  {{ fail "configyaml must be a map" }}
+{{- else }}
+  {{- .Values.starrocksFESpec.config | nindent 2 }}
 {{- end }}
 {{- end }}
 
 {{- define "starrockscluster.cn.config" -}}
 cn.conf: |
-{{- if .Values.starrocksCnSpec.config }}
-{{- .Values.starrocksCnSpec.config | nindent 2 }}
+{{- if and .Values.starrocksCnSpec.configyaml (kindIs "map" .Values.starrocksCnSpec.configyaml) }}
+  {{- range $key, $value := .Values.starrocksCnSpec.configyaml }}
+    {{ $key }} = {{ $value }}
+  {{- end }}
+{{- else if .Values.starrocksCnSpec.configyaml }}
+  {{ fail "configyaml must be a map" }}
+{{- else }}
+  {{- .Values.starrocksCnSpec.config | nindent 2 }}
 {{- end }}
 {{- end }}
 
 {{- define "starrocksclster.be.config" -}}
 be.conf: |
-{{- if .Values.starrocksBeSpec.config }}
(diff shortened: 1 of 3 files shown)


yandongxiao:
‌‌‌‌‌‌‌I think there is a backward compatibility issue here. That is, if the user has set a custom config by `config` and then upgrades to use the latest helm chart, because the value of `configyaml` is always not empty, it will cause the user to use the default content of `configyaml` as the configuration for FE/BE/CN.

yandongxiao:
Looking at your code, you directly modified `helm-charts/charts/kube-starrocks/values.yaml`, should modify `helm-charts/charts/kube-starrocks/charts/starrocks/values.yaml`.

In `scripts` directory, run `bash create-parent-chart-values.sh` to update the values.yaml file of the parent
chart( kube-starrocks chart).

yandongxiao:
‌Can we merge the contents of `config` and `configyaml` to generate the final result? This way, there will only be one set of configurations in values.yaml. At the same time, allow users to add incremental configurations through `configyaml`.

dengliu:
> ‌‌‌‌‌‌‌I think there is a backward compatibility issue here. That is, if the user has set a custom config by `config` and then upgrades to use the latest helm chart, because the value of `configyaml` is always not empty, it will cause the user to use the default content of `configyaml` as the configuration for FE/BE/CN.

I can set configyaml to `{}` to address this issue. would this solution OK with you?

dengliu:
> update



> > ‌Can we merge the contents of `config` and `configyaml` to generate the final result? This way, there will only be one set of configurations in values.yaml. At the same time, allow users to add incremental configurations through `configyaml`.

I am afraid this hybrid mode will cause more confusion, and increase the chance of making mistakes

yandongxiao:
> > ‌‌‌‌‌‌‌I think there is a backward compatibility issue here. That is, if the user has set a custom config by `config` and then upgrades to use the latest helm chart, because the value of `configyaml` is always not empty, it will cause the user to use the default content of `configyaml` as the configuration for FE/BE/CN.
> 
> I can set configyaml to `{}` to address this issue. would this solution OK with you?

This is LGTM. The method I mentioned seems quite difficult to implement, and splitting the configuration into two blocks doesn't seem like a good choice.

yandongxiao:
> > ‌Can we merge the contents of `config` and `configyaml` to generate the final result? This way, there will only be one set of configurations in values.yaml. At the same time, allow users to add incremental configurations through `configyaml`.
> 
> if both config and configyaml has the same key, which value shall be overriden? I am afraid this hybrid mode will cause more confusion, and increase the chance of making mistakes

Agree with you

kevincai (approved):

PR #583, opened 2024-08-29
[Chore] Prepare for v1.9.8

yandongxiao:


index.yaml
```diff
@@ -1,6 +1,40 @@
 apiVersion: v1
 entries:
   kube-starrocks:
+  - apiVersion: v2
+    appVersion: 3.2-latest
```
kevincai, line 5: change to 3.3-latest, since we already have 3.3 published for months

yandongxiao: OK

yandongxiao: done

kevincai (approved):

PR #621, opened 2025-02-18
Feature/support disaster recovery

yandongxiao:
# Description

From StarRocks 3.4.1, the shared-data mode cluster supports disaster recovery. The Operator can configure the disaster
recovery for the shared-data mode cluster to ensure the data security and high availability.

# Checklist

  file of starrocks chart.
  chart( kube-starrocks chart).


config/crd/bases/starrocks.com_starrocksclusters.yaml
```diff
@@ -1067,6 +1067,21 @@ spec:
                           type: string
                       type: object
                     type: array
+                  disasterRecovery:
```
kevincai, line 1070: should be a cluster-wise property, no need to break down to sub controllers?

kevincai (commented):
overall looks good to me, check if we can move the disaster recovery setting and status to crd level instead of each subcontroller.

kevincai (approved):

Issue #625, opened 2025-02-22
Support user-specified service labels


dd-willgan:
## Describe the current behavior

Only annotations can be user-specified

## Describe the enhancement

Support user-specified service labels in the operator and Helm chart for e.g. starrocks-fe-search k8s service

## Additional context

At my company labels are used to automatically generate URLs for the service, which we want to use

dd-willgan:
If this sounds like an okay enhancement, I can look into implementing it!

yandongxiao:
This is a good enhancement, welcome to contribute code. It should be noted that for each component, the operator will generate two services, for example, kube-starrocks-fe-service and kube-starrocks-fe-search. The former is exposed to users for use, while the latter is used internally by StarRocks through headless service to give each Pod a network identity. I think it should only be necessary to support adding labels for `*-service`.

dd-willgan:
Thanks @yandongxiao ! Took a stab at a PR adding only to -service and not -search

yandongxiao:
From the current main branch perspective, the internal service `-search` will copy the label of external service `-service`, so their labels always remain the same.

PR #632, opened 2025-03-18
Support service labels

dd-willgan:
# Description

Currently adding K8s service annotations are supported but not labels. Support service labels through the following changes:

- Update CRD definitions in `config/crd/bases/` and `deploy/`
- Update Helm chart templates and values for `helm-charts/charts/kube-starrocks/` and `helm-charts/charts/warehouse/`
- Update `pkg/apis/starrocks/v1/`
- Update `pkg/common/resource_utils/service` and `pkg/k8sutils/templates/service/spec` to handle labels
  - Still support default labels generated by operator
- Update subcontrollers so that external service has labels but internal labels don't

# Related Issue(s)

https://github.com/StarRocks/starrocks-kubernetes-operator/issues/625

# Checklist

- [X] run `make generate` to generate the code.
- [X] run `make test` to run UT.
- [X] run `make manifests` to update the yaml files of CRD.

- [X] make sure you have updated the [values.yaml](../helm-charts/charts/kube-starrocks/charts/starrocks/values.yaml)
  file of starrocks chart.
- [X] In `scripts` directory, run `bash create-parent-chart-values.sh` to update the values.yaml file of the parent
  chart( kube-starrocks chart).

### DIFF
--- a/helm-charts/charts/kube-starrocks/charts/starrocks/templates/starrockscluster.yaml
+++ b/helm-charts/charts/kube-starrocks/charts/starrocks/templates/starrockscluster.yaml
@@ -68,6 +68,10 @@ spec:
         {{- if .Values.starrocksFESpec.service.annotations}}
         {{- toYaml .Values.starrocksFESpec.service.annotations | nindent 8 }}
         {{- end }}
+      labels:
+        {{- if .Values.starrocksFESpec.service.labels}}
+        {{- toYaml .Values.starrocksFESpec.service.labels | nindent 8 }}
+        {{- end }}
       {{- end }}
     {{- end }}
     annotations:
@@ -324,6 +328,10 @@ spec:
         {{- end }}
         {{- if .Values.starrocksBeSpec.service.annotations }}
         {{- toYaml .Values.starrocksBeSpec.service.annotations | nindent 8 }}
+        {{- end }}
+      labels:
+        {{- if .Values.starrocksBeSpec.service.labels}}
+        {{- toYaml .Values.starrocksBeSpec.service.labels | nindent 8 }}
         {{- end }}
       {{- end }}
     {{- end }}
@@ -750,6 +758,10 @@ spec:
         {{- if .Values.starrocksCnSpec.service.annotations }}
         {{- toYaml .Values.starrocksCnSpec.service.annotations | nindent 8 }}
         {{- end }}
+      labels:
+        {{- if .Values.starrocksCnSpec.service.labels}}
+        {{- toYaml .Values.starrocksCnSpec.service.labels | nindent 8 }}
+        {{- end }}
       {{- end }}
     {{- end }}
     annotations:
@@ -874,6 +886,10 @@ spec:
       {{- if .Values.starrocksFeProxySpec.service.annotations }}
       annotations:
         {{- toYaml .Values.starrocksFeProxySpec.service.annotations | nindent 8 }}
+      {{- end }}
+      {{- if .Values.starrocksFeProxySpec.service.labels}}
+      labels:
+        {{- toYaml .Values.starrocksFeProxySpec.service.labels | nindent 8 }}
       {{- end }}
       {{- if and (eq "LoadBalancer" .Values.starrocksFeProxySpec.service.type) .Values.starrocksFeProxySpec.service.loadbalancerIP }}
       loadBalancerIP: {{ .Values.starrocksFeProxySpec.service.loadbalancerIP }}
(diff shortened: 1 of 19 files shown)


pkg/k8sutils/templates/service/spec.go
```diff
@@ -20,13 +20,15 @@ import (
...
-func MakeSearchService(serviceName string, externalService *corev1.Service, ports []corev1.ServicePort) *corev1.Service {
+func MakeSearchService(serviceName string, externalService *corev1.Service, ports []corev1.ServicePort,
+	labels map[string]string) *corev1.Service {
 	searchSvc := &corev1.Service{}
 	externalService.ObjectMeta.DeepCopyInto(&searchSvc.ObjectMeta)
 	// some service annotations can only be used when `type` is 'LoadBalancer', e.g. service.beta.kubernetes.io/load-balancer-source-ranges
 	// we do not need to annotations for search service
 	searchSvc.Annotations = nil
 	searchSvc.Name = serviceName
+	searchSvc.Labels = labels
```
yandongxiao, line 31: It looks like the DeepCopyInto method above has already copied all labels from the External service, so there should be no need to reassign labels.

helm-charts/charts/kube-starrocks/charts/starrocks/templates/starrockscluster.yaml
```diff
@@ -68,6 +68,10 @@ spec:
         {{- if .Values.starrocksFESpec.service.annotations}}
         {{- toYaml .Values.starrocksFESpec.service.annotations | nindent 8 }}
         {{- end }}
+      labels:
+        {{- if .Values.starrocksFESpec.service.labels}}
```
yandongxiao, line 72: I think, in order to be consistent with other fields, the if statement should be placed above the labels.

pkg/subcontrollers/be/be_controller.go
```diff
@@ -106,10 +106,11 @@ func (be *BeController) SyncCluster(ctx context.Context, src *srapi.StarRocksClu
 	// add query port from fe config.
 	config[rutils.QUERY_PORT] = strconv.FormatInt(int64(rutils.GetPort(feConfig, rutils.QUERY_PORT)), 10)
 	// generate new be external service.
+	defaultLabels := load.Labels(src.Name, beSpec)
 	externalsvc := rutils.BuildExternalService(object.NewFromCluster(src),
-		beSpec, config, load.Selector(src.Name, beSpec), load.Labels(src.Name, beSpec))
+		beSpec, config, load.Selector(src.Name, beSpec), defaultLabels)
 	// generate internal fe service, update the status of cn on src.
-	internalService := be.generateInternalService(ctx, src, &externalsvc, config)
+	internalService := be.generateInternalService(ctx, src, &externalsvc, config, defaultLabels)
```
yandongxiao, line 113: I think the defaultLabels parameter is unnecessary. The internal service will automatically copy the labels in the External Service.

yandongxiao (changes requested):

helm-charts/charts/kube-starrocks/charts/starrocks/templates/starrockscluster.yaml
dd-willgan: Ah yes, updated

pkg/subcontrollers/be/be_controller.go
dd-willgan: Hmm for my use case I think I only want the external service to have the labels, not the internal service. That's what you mentioned in this [comment](https://github.com/StarRocks/starrocks-kubernetes-operator/issues/625#issuecomment-2677286562) from a few weeks ago as well?

dd-willgan: Ah sorry saw you commented in that issue. Is it possible to have differentiated labels? For more context, the labels are used to by some of my company's internal tooling to generate a URL for the service. If I have labels on both the internal and external service, the URL could route to the either the internal or external.

yandongxiao: I got it. There is indeed no need to keep the Label consistent between Internal and external service, and the Annotation field also only works on external service. Can we add some comments in `values.yaml` and `labels` fields to indicate that they only apply to external service? If possible, can we also update the comments for `annotations`? The way `annotations` is currently handled in this aspect is not very good.

pkg/k8sutils/templates/service/spec.go
dd-willgan: Done

pkg/subcontrollers/be/be_controller.go
dd-willgan: Updated the comments in the CRD and generated files, `values.yaml`'s, and in `spec.go`

yandongxiao (approved):

yandongxiao:
@dd-willgan  Thank you for your work!

PR #679, opened 2025-07-31
[BugFix] Rename the short name for warehouse

yandongxiao:
# Description

Rename the short name for warehouse, so that we can do like this:
```
➜  kubectl get warehouses
NAME   STATUS    REASON
wh-1   running
```
# Checklist


kevincai (commented):
concerns about the compatibility, will this require use to delete the crd and re apply the definition?

yandongxiao:
> concerns about the compatibility, will this require use to delete the crd and re apply the definition?

just need to reapply the CRD

PR #684, opened 2025-08-01
[Doc] Replace automatic scaling guide with comprehensive HPA documentation

yandongxiao:
# Description

- Remove outdated automatic_scaling_for_cn_nodes_howto.md
- Add hpa_dynamic_scaling_with_helm_howto.md with detailed HPA implementation guide
- Document both integrated cluster and separate warehouse deployment methods
- Include v1.10.2+ feature improvements and complete configuration examples


doc/automatic_scaling_for_cn_nodes_howto.md
kevincai, line 1: maybe need to keep this doc and indicate that it is only valid with operator <1.10.2?

yandongxiao: The reasons I want to delete it are: 1. The content of this file overlaps significantly with that of the new file; 2. We should guide users to use the latest configuration. If the old configuration still exists, users may still use the older one if they are not careful.

yandongxiao: ok

kevincai (approved):

Issue #649, opened 2025-05-19
安装helm chart报错，提示runAsNonRoot 值为null


owedocker:
## Please complete the following information

- Operator Version: [1.10.2]
- Chart Name: [kube-starrocks]
- Chart Version [1.10.2]

使用helm 部署kube-starrocks，报错
![Image](https://github.com/user-attachments/assets/f8850e43-b42e-4d8f-b084-c5ae038763da)

另外进入子chart，starrocks，同样helm install --dry-run 生成出来的yaml文件没值。导致部署失败，
以下是我认为可能的原因

因为Helm的模板语言中，如果变量的值为false，那么if条件判断会认为是false，从而不会进入该条件分支。例如，在模板中：

{{- define "starrockscluster.fe.runAsNonRoot" -}}
{{- if .Values.starrocksFESpec.runAsNonRoot }}
{{- .Values.starrocksFESpec.runAsNonRoot }}
{{- else if ... }}
{{- end }}

此时，当.Values.starrocksFESpec.runAsNonRoot为false时，if条件不成立，所以会进入else if部分。如果else if部分没有定义，则整个define会返回空，导致最终生成的YAML中runAsNonRoot字段为null，因为模板没有输出任何内容。

建议修改为以下：
{{- define "starrockscluster.fe.runAsNonRoot" -}}
{{- .Values.starrocksFESpec.runAsNonRoot | default .Values.starrocksCluster.componentValues.runAsNonRoot }}
{{- end }}

owedocker:
完整修复如下：

{{/*
Common labels
*/}}
{{- define "starrockscluster.labels" -}}
app.kubernetes.io/instance: {{ .Release.Name }}
app.kubernetes.io/managed-by: {{ .Release.Service }}
{{- end }}

{{/*
initpassword secret name
*/}}

{{- define "starrockscluster.initpassword.secret.name" -}}
{{ default (print (include "starrockscluster.name" .) "-credential") .Values.initPassword.passwordSecret }}
{{- end }}

{{/*
starrockscluster
*/}}

{{- define "starrockscluster.name" -}}
{{ default (default .Chart.Name .Values.nameOverride) .Values.starrocksCluster.name }}
{{- end }}

{{- define "starrockscluster.namespace" -}}
{{ default .Release.Namespace .Values.starrocksCluster.namespace }}
{{- end }}

{{- define "starrockscluster.fe.name" -}}
[...]

PR #693, opened 2025-09-03
fix: add default value false with runAsNonRoot

Sn0rt:
# Description

Please provide a detailed description of the changes you have made. Include the motivation for these changes, and any
additional context that may be important.
> the PR should include helm chart changes and operator changes.

# Related Issue(s)

https://github.com/StarRocks/starrocks-kubernetes-operator/issues/649

Please list any related issues and link them here.

# Checklist

- [ ] make sure you have updated the [values.yaml](../helm-charts/charts/kube-starrocks/charts/starrocks/values.yaml)
  file of starrocks chart.
- [ ] In `scripts` directory, run `bash create-parent-chart-values.sh` to update the values.yaml file of the parent
  chart( kube-starrocks chart).

### DIFF
--- a/helm-charts/charts/kube-starrocks/charts/starrocks/templates/_helpers.tpl
+++ b/helm-charts/charts/kube-starrocks/charts/starrocks/templates/_helpers.tpl
@@ -618,6 +618,8 @@ Get the value of the runAsNonRoot field in the starrocksFESpec
 {{- .Values.starrocksFESpec.runAsNonRoot -}}
 {{- else if .Values.starrocksCluster.componentValues.runAsNonRoot -}}
 {{- .Values.starrocksCluster.componentValues.runAsNonRoot -}}
+{{- else -}}
+false
 {{- end -}}
 {{- end -}}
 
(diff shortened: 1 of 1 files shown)


helm-charts/charts/kube-starrocks/charts/starrocks/templates/_helpers.tpl
```diff
@@ -614,32 +614,38 @@ Get the value of the topologySpreadConstraints field in the starrocksCnSpec
 Get the value of the runAsNonRoot field in the starrocksFESpec
 */}}
 {{- define "starrockscluster.fe.runAsNonRoot" -}}
-{{- if .Values.starrocksFESpec.runAsNonRoot -}}
+{{- if hasKey .Values.starrocksFESpec "runAsNonRoot" -}}
```
yandongxiao, line 617: It may be a k8s version issue that prevents me from reproducing the problem mentioned in the issue. My only question is why hasKey is needed here? Can the original format still be used? This is because a large number of similar expressions from before are used in this file.

yandongxiao: So can you please use the old way to detect the condition, not the `hasKey `.

yandongxiao (approved):

PR #709, opened 2025-10-16
[Doc] Add doc/scale_in_shared_nothing_cluster_howto.md

yandongxiao:


doc/scale_in_shared_nothing_cluster_howto.md
```diff
@@ -0,0 +1,135 @@
...
+   ```
+
+2. Drop the FE node from the StarRocks cluster.
+   ```sql
+   mysql> ALTER SYSTEM DROP FOLLOWER "kube-starrocks-fe-2.kube-starrocks-fe-search.default.svc.cluster.local:9010";
+   Query OK, 0 rows affected (0.22 sec)
+   ```
+
+3. Adjust the `replicas` field to a smaller number, e.g. 3-->2.
+4. Repeat the above steps to remove other FE nodes until the desired number of FE nodes is reached. Note that it is not
```
kevincai: It is not a good example to scale-in FE from 3->1, it will fail for sure, because the BDBJE HAGroup can't change from HA mode to single node mode automatically. An example of scale in from 5->3, is appropriate.  And warn user don't try to scale from 3->1, if insist to scale in from 3->1, a metadata recovery process will be necessary for the case: https://docs.starrocks.io/docs/administration/Meta_recovery/#recover-metadata-on-a-new-fe-node-with-metadata-backup

PR #734, opened 2026-01-06
[Feature] Support adding an observer role in FE to improve query load balancing and scale read workloads

zhaohehuhu:
# Description

Support adding a read-only observer role in FE via the K8s Operator to scale read capacity and improve query load balancing.

associated with this PR(https://github.com/StarRocks/starrocks/pull/67778)

# Related Issue(s)

No.

# Checklist

  file of starrocks chart.
  chart( kube-starrocks chart).


kevincai:
don't try to use FE_OBSERVER_NUMBER to hack the current fe_entry_point.sh

It is so dangerous that a single improper env var update can ruin the entire cluster.

zhaohehuhu:
> don't try to use FE_OBSERVER_NUMBER to hack the current fe_entry_point.sh
> 
> It is so dangerous that a single improper env var update can ruin the entire cluster.

Got it. In that case, we may need to add a new FE observer component.

zhaohehuhu:
> This is much needed, thank you!

I still need some time to test this feature. Thanks!

zhaohehuhu:
@kevincai @yandongxiao plz help review. Thanks!

helm-charts/charts/kube-starrocks/charts/starrocks/templates/starrockscluster.yaml
```diff
@@ -284,6 +284,181 @@ spec:
...
+    shareProcessNamespace: {{ .Values.starrocksFeObserverSpec.shareProcessNamespace }}
+    {{- end }}
+    image: "{{ .Values.starrocksFeObserverSpec.image.repository }}:{{ default .Values.starrocksCluster.componentValues.image.tag .Values.starrocksFeObserverSpec.image.tag }}"
+    replicas: {{ .Values.starrocksFeObserverSpec.replicas }}
+    imagePullPolicy: {{ .Values.starrocksFeObserverSpec.imagePullPolicy }}
+    {{- if .Values.starrocksFeObserverSpec.resources }}
+      {{- toYaml .Values.starrocksFeObserverSpec.resources | nindent 4 }}
+    {{- end }}
+    {{- if .Values.starrocksFeObserverSpec.service.type }}
+    service:
```
emanuilov, line 303: Do we need a separate observer service, can't we share the general one?

pkg/subcontrollers/feobserver/feobserver_config.go
```diff
@@ -0,0 +1,19 @@
+package feobserver
```
emanuilov, line 1: Why is this package necessary, can't we reuse the general one, only difference here should be the config, correct?

helm-charts/charts/kube-starrocks/charts/starrocks/values.yaml
```diff
@@ -514,6 +514,210 @@ starrocksFESpec:
   # in the same pod, and the first process in each container will not be assigned PID 1.
   shareProcessNamespace:
 
+# spec to deploy fe observer.
+starrocksFeObserverSpec:
```
emanuilov, line 518: Am I missing something, isn't this PR supposed to add the functionallity for an observer pod/statefulset to be added, I do not see any custom config/boot behavior configuring the FE pod to be an observer

Also why are the starrocksFeObserverSpec.config and .configyaml fields missing here

helm-charts/charts/kube-starrocks/charts/starrocks/templates/starrockscluster.yaml
```diff
@@ -284,6 +284,181 @@ spec:
     persistentVolumeClaimRetentionPolicy:
       {{- toYaml .Values.starrocksFESpec.persistentVolumeClaimRetentionPolicy | nindent 6 }}
     {{- end }}
+  {{- if .Values.starrocksFeObserverSpec.enabled }}
+  starRocksFeObserverSpec:
```
emanuilov, line 288: Do we need most of the fields here, can't we reuse the general FE ones and provide a custom config or smth?

helm-charts/charts/kube-starrocks/charts/starrocks/templates/starrockscluster.yaml
```diff
@@ -284,6 +284,181 @@ spec:
...
+    {{- if .Values.starrocksFeObserverSpec.initContainers }}
+    initContainers:
+      {{- toYaml .Values.starrocksFeObserverSpec.initContainers | nindent 6 }}
+    {{- end }}
+    {{- if .Values.starrocksFeObserverSpec.shareProcessNamespace }}
+    shareProcessNamespace: {{ .Values.starrocksFeObserverSpec.shareProcessNamespace }}
+    {{- end }}
+    image: "{{ .Values.starrocksFeObserverSpec.image.repository }}:{{ default .Values.starrocksCluster.componentValues.image.tag .Values.starrocksFeObserverSpec.image.tag }}"
```
emanuilov, line 296: Are regular FE nodes required for the FE observer nodes to work? If yes, even if no, wouldn't it be simpler to rewrite this MR so that:
1. We add a starrocksFESpec.observerSpec with .enabled and .autoScalingPolicy (same as the CN spec) this can have HPA settings, and simply use the --helper command (according to Rocky pretty much all is needed to start an observer) instead of the generic one

All the other fields can be reused from the generic FE resources, no separate service would make using this even simpler.

zhaohehuhu: Thanks for the input. This looks much simpler and I’ll chat with the community member about it. @emanuilov

emanuilov (approved):

kevincai (dismissed):

yandongxiao:
Thanks for the contribution. The overall mechanism — a dedicated observer workload plus the minimal `fe_entrypoint.sh` change in StarRocks/starrocks#67778 — looks like the right direction. However, I have four concerns. The first one is blocking.

## 1. Observer pods have no resolvable FQDN, and no Service routes to them (blocking)

`MakeStatefulset` initially sets `ServiceName` to `<cluster>-fe-observer-search` (the new case in `service.SearchServiceName`), but `SyncCluster` immediately overrides it:

```go
expectSts.Spec.ServiceName = feSearchServiceName(src.Name) // "<cluster>-fe-search"
```

That service does exist (created by the FE controller), but its selector is built from `load.Selector(src.Name, feSpec)`, i.e. `{app.starrocks.ownerreference/name: <cluster>-fe, app.kubernetes.io/component: fe}`, while observer pods carry `{<cluster>-fe-observer, fe-observer}`. The selector never matches, so observer pods never appear in the Endpoints of `<cluster>-fe-search`, and cluster DNS never publishes A records for `<pod>.<cluster>-fe-search.<ns>.svc.cluster.local`.

The pod can still resolve its **own** FQDN locally (kubelet writes hostname/subdomain into `/etc/hosts`), so `ALTER SYSTEM ADD OBSERVER "$MYSELF:$EDIT_LOG_PORT"` is actually sent — but the FE leader cannot resolve that FQDN, BDBJE cannot connect back, and the entrypoint times out and exits → CrashLoopBackOff. Since the operator always sets `HOST_TYPE=FQDN`, the feature cannot work as currently implemented.

Additionally, because the dedicated observer services were removed, **no Service selects observer pods at all** — there is no entry point for clients to reach observers, which seems at odds with the stated goal of query load balancing.

Suggestion: keep a dedicated headless search service for the observer StatefulSet (selector = observer labels), which also makes the currently-dead `*StarRocksFeObserverSpec` case in `service.SearchServiceName` meaningful again. It is also worth clarifying how clients are expected to reach observers (e.g. a dedicated external service).

## 2. `deleteLegacyObserverServices` should be removed

It deletes `<cluster>-fe-observer-search` / `<cluster>-fe-observer-service`, which were only ever created by earlier revisions of this PR and never released — no real deployment can have these resources.

It also runs on the **enabled** path on every reconcile, and `clearObserverResources` runs on the **disabled** path (the default) on every reconcile of every cluster. `DeleteService`/`DeleteStatefulset` log `"delete service/statefulset from kubernetes"` *before* the existence check, so every reconcile of every cluster emits misleading delete logs plus pointless GETs. Consider gating the disabled-path cleanup on `src.Status.StarRocksFeObserverStatus != nil`.

(Minor: the `!apierrors.IsNotFound(err)` checks are redundant — `DeleteService`/`DeleteStatefulset` already return nil on NotFound.)

## 3. No deregistration on scale-down / disable

Neither the operator nor `fe_entrypoint.sh` ever runs `ALTER SYSTEM DROP OBSERVER`. Scaling `observerNumber` down or setting `enabled: false` leaves permanent `Alive=false` entries in `SHOW FRONTENDS`. This matches the existing FE follower convention (manual DROP), but observers are precisely the role intended to be scaled dynamically, so dead entries will accumulate much faster. This should at least be documented as a known limitation, with manual cleanup instructions.

## 4. Old image + `observerSpec.enabled=true` silently adds FOLLOWERs (dangerous)

`IS_FE_OBSERVER` is only understood by `fe_entrypoint.sh` after StarRocks/starrocks#67778. With any older FE image, the env var is ignored and the script falls through to `ALTER SYSTEM ADD FOLLOWER` — the observer replicas join as **voting followers** and silently change the BDBJE quorum. This is exactly the failure mode raised earlier in this thread regarding env-var-driven behavior.

Suggestion: add an operator-side image version guard (`IsLowerThanAny` is already used this way for the BE preStop path) and document the minimum required FE image version for `observerSpec`.

pkg/subcontrollers/feobserver/feobserver_controller.go
yandongxiao, line 1: no drop observer logic

yandongxiao (changes requested):

zhaohehuhu:
@yandongxiao Thanks for your comments. I will make some improvements.

pkg/subcontrollers/feobserver/feobserver_controller.go
zhaohehuhu: I’d prefer to add that logic in a separate PR to keep the scope of this PR small. Would that be okay? @yandongxiao

zhaohehuhu:
@yandongxiao  plz help review. Thanks!

vikasattiguppa:
@yandongxiao can you please review these changes i'm also interested in getting the capability for our starrocks deployments.

kevincai (dismissed):

digitalpoetry:
Related issue https://github.com/StarRocks/starrocks-kubernetes-operator/issues/767

digitalpoetry:
@yandongxiao any update, please?

pkg/apis/starrocks/v1/starrockscluster_types.go
```diff
@@ -157,10 +153,29 @@ type StarRocksClusterStatus struct {
...
+	// ObserverNumber is the number of desired FE observer Pods.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=1
+	// +optional
+	ObserverNumber *int32 `json:"observerNumber,omitempty"`
+
+	ComponentSpec StarRocksComponentSpec `json:"-"`
+
+	FeEnvVars []corev1.EnvVar `json:"-"`
 }
```
yandongxiao, line 179: I don't think ComponentSpec and FeEnvVars need to be defined under StarRocksFeObserverSpec. It seems a bit confusing, as if users can configure these two fields, or as if the API Server will store the values of these two fields.

pkg/subcontrollers/feobserver/feobserver_controller.go
```diff
@@ -0,0 +1,236 @@
...
+	var st appsv1.StatefulSet
+	if err := fc.Client.Get(context.Background(), types.NamespacedName{Namespace: src.Namespace, Name: statefulSetName}, &st); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// ClearCluster clears resource about fe observer.
+func (fc *FeObserverController) ClearCluster(ctx context.Context, src *srapi.StarRocksCluster) error {
```
yandongxiao, line 181: Missing logic to call ClearCluster. You can refer to the calling logic of ClearCluster for be and cn

pkg/k8sutils/templates/service/spec.go
```diff
@@ -63,6 +65,8 @@ func ExternalServiceName(clusterName string, spec v1.SpecInterface) string {
 	switch spec.(type) {
 	case *v1.StarRocksFeSpec:
 		return clusterName + "-" + v1.DEFAULT_FE + "-service"
+	case *v1.StarRocksFeObserverSpec:
+		return clusterName + "-" + v1.DEFAULT_FE_OBSERVER + "-service"
```
yandongxiao: If FE and FE Observer aren't under the same Kubernetes service, then we won't be able to expose a unified entry point. This causes trouble for StarRocks users. 
So, I do not think we need an external service for observer.

pkg/subcontrollers/feobserver/feobserver_controller.go
```diff
@@ -0,0 +1,236 @@
...
+
+	if err := subcontrollers.UpdateStatus(&fs.StarRocksComponentStatus, fc.Client,
+		src.Namespace, statefulSetName, pod.Labels(src.Name, observerSpec), subcontrollers.StatefulSetLoadType); err != nil {
+		return err
+	}
+
+	var st appsv1.StatefulSet
+	if err := fc.Client.Get(context.Background(), types.NamespacedName{Namespace: src.Namespace, Name: statefulSetName}, &st); err != nil {
+		return err
+	}
```
yandongxiao: There is a scenario where the FE Observer StatefulSet does not exist. When a cluster is initially created, if the FE Statefulset's Pods are not yet Ready, it will result in the FE Observer StatefulSet not being created.
Please see BE/CN's logic.

pkg/subcontrollers/feobserver/feobserver_controller.go
```diff
@@ -0,0 +1,236 @@
...
+			Phase: srapi.ComponentReconciling,
+		},
+	}
+
+	if src.Status.StarRocksFeObserverStatus != nil {
+		fs = src.Status.StarRocksFeObserverStatus.DeepCopy()
+	}
+
+	src.Status.StarRocksFeObserverStatus = fs
+	fs.ServiceName = service.ExternalServiceName(src.Name, feSpec)
```
yandongxiao: feSpec is a wrong parameter?

yandongxiao (changes requested):

pkg/apis/starrocks/v1/starrockscluster_types.go
zhaohehuhu: Thanks for reviewing. After removing ComponentSpec and FeEnvVars from StarRocksFeObserverSpec, it is no longer a full component spec and should not implement SpecInterface. As a result, helpers such as MakeStatefulset, which currently accept SpecInterface, can no longer take StarRocksFeObserverSpec directly. We need to refactor many methods, which should add some extra work. @yandongxiao

pkg/subcontrollers/feobserver/feobserver_controller.go
zhaohehuhu: Thanks. Fixed.

pkg/subcontrollers/feobserver/feobserver_controller.go
zhaohehuhu: Thanks. Fixed.

pkg/k8sutils/templates/service/spec.go
zhaohehuhu: Thanks. Fixed.

pkg/subcontrollers/feobserver/feobserver_controller.go
zhaohehuhu: Thanks. Fixed.

pkg/apis/starrocks/v1/starrockscluster_types.go
zhaohehuhu: Can we mark both two fields as unexported ? @yandongxiao

kevincai:
@codex review

kevincai:
@codex review

digitalpoetry:
@yandongxiao @kevincai Any updates please? My team is keenly awaiting this change

kevincai (approved):

PR #743, opened 2026-03-19
feat (helm): add ingress template

cgpoh:
# Description

Since we are not exposing the service via a LoadBalancer and instead rely on an Ingress controller to route traffic to the StarRocks service, an Ingress template is required.

> This PR includes helm chart changes only. No operator changes are included.

# Related Issue(s)

N/A

# Checklist

- N/A - This PR includes Helm chart changes only. No operator code changes.

  file of starrocks chart.
  chart( kube-starrocks chart).

### DIFF
--- a/helm-charts/charts/kube-starrocks/values.yaml
+++ b/helm-charts/charts/kube-starrocks/values.yaml
@@ -221,17 +221,19 @@ starrocks:
         #      - "all"
         params: {}
   
-  # Configure Ingress to expose FE TCP endpoint.
+  # Configure Ingress to expose the FE MySQL endpoint.
   ingress:
     enabled: false
+    # Override ingress resource name. If empty, defaults to <cluster-name>-mysql-ingress.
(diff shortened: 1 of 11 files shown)


helm-charts/charts/kube-starrocks/values.yaml
```diff
@@ -221,6 +221,30 @@ starrocks:
         #      - "all"
         params: {}
   
+  # Configure Ingress to expose FE TCP endpoint.
+  ingress:
```
kevincai, line 225: it could be multiple ingresses in front of the services for different tcp port. better have a name for the ingress. Also this hardcoded `servicePort:9030` is misleading, if it is only added for mysql protocol, can make this ingress configuration as specific as possible.

cgpoh: Thanks @kevincai , I've pushed my changes to address your comments.

PR #748, opened 2026-04-16
[Feature] Add global.image.registry support for custom registry prefix

Gui13:
# Description

This allows to use private registries without having to override the repository. This is how public charts usually work: split between registry, repository and image tag. Bitnami used to work like this.

Adds a global.image.registry value (default: "") that is prepended to all component image repositories. Each component (FE, BE, CN, FE proxy) also exposes its own image.registry field to allow per-component overrides; the component-level value takes priority over the global one.

New helpers (starrockscluster.fe/be/cn/feproxy.image) build the full image reference in one place, keeping templates clean. The init-password job inherits the FE registry automatically.

Priority order (ascending, last wins):
  global.image.registry
  < starrocksCluster.componentValues.image.registry
  < starrocks{FE,BE,CN}Spec.image.registry

The feproxy component is excluded from componentValues (as it already was for image.tag) and resolves directly from its own registry or global.image.registry.

This commit was partially helped with Claude, I added it in the commit Co-Authors.

# Related Issue(s)

None

# Checklist

  file of starrocks chart.
  chart( kube-starrocks chart).

### DIFF
--- a/helm-charts/charts/kube-starrocks/charts/starrocks/templates/_helpers.tpl
+++ b/helm-charts/charts/kube-starrocks/charts/starrocks/templates/_helpers.tpl
@@ -716,6 +716,69 @@ Get the value of tag field in the starrocksCnSpec
 {{- end -}}
 
 {{/*
+Build the full FE image reference, prepending the registry when set.
+Priority (highest wins): component image.registry > componentValues.image.registry > global.image.registry.
+*/}}
+{{- define "starrockscluster.fe.image" -}}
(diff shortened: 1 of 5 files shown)


Gui13:
Please note that I have an helm unittest (https://github.com/helm-unittest/) file that tests the combinations of global.image.registry / starrocksComponents.image.registry and individual registry values to validate that the proper overrides are working as intended. 

I didn't commit the file because this would enlarge the merge request substantially :
- add unittest/ folder & the unittest file
- need to add a new github action to pass theses unittests (otherwise they are not useful per se)
- need to add a .helmignore file to ignore packaging of this unittest/ folder

PR #762, opened 2026-06-01
[Feature] Add VolumeAttributesClass support to StorageVolume

### LINKED ISSUE #749: Add support Volume Attributes Classes to modify the volume type, IOPS, or throughput of your EBS volumes on the fly
## Enhancement
Add support Volume Attributes Classes `kind: VolumeAttributesClass` to modify the volume type, IOPS, or throughput of your EBS volumes on the fly in helm chart.

Link:
https://aws.amazon.com/blogs/containers/modify-amazon-ebs-volumes-on-kubernetes-with-volume-attributes-classes/

AdheipSingh:
## Summary
- Adds optional `volumeAttributeClassName` field to `StorageVolume`, enabling dynamic modification of EBS volume type, IOPS, and throughput via Kubernetes VolumeAttributesClass (k8s 1.31+)
- Since the vendored k8s API (v0.26.1) lacks `VolumeAttributeClassName` on PVC spec, the operator patches individual PVCs at runtime using unstructured merge patches after StatefulSet reconciliation
- Helm chart support added for all components (FE, BE, CN) and the warehouse chart

Closes #749

## Changes
- **API**: New optional `volumeAttributeClassName` field on `StorageVolume`
- **Operator**: `PatchPVCVolumeAttributeClass` utility patches existing PVCs after StatefulSet apply in FE, BE, and CN controllers
- **CRDs**: Updated cluster and warehouse CRD manifests
- **Helm**: `volumeAttributeClassName` value added to FE, BE, CN storageSpec in both kube-starrocks and warehouse charts


TiansuYu:
Hey, this is super useful for our use cases as well. Looking forward for it to be released. 

Just want to point out: The PVC field is `volumeAttributesClassName` (plural *Attributes*), but the PR uses singular `volumeAttributeClassName` everywhere, including the patch:

```go
`{"spec":{"volumeAttributeClassName":"%s"}}`
```

That key doesn't match the real PVC spec, so the patch won't change IOPS/throughput.

Docs (note the plural): https://kubernetes.io/docs/reference/kubernetes-api/config-and-storage-resources/persistent-volume-claim-v1/#PersistentVolumeClaimSpec

Please double check on this. :)

TiansuYu:
Another comment I want to make is: Is it possible to bind different VolumeAttributesClass to different types of disks (meaning `VolumeAttributesClassName` is not bind on top level, but can be declared per `data`, `spill` and `log`). For large clusters, I think this would save some bucks, especially for `log` disks, which doesn't need that much IO throughput. 

Not necessarily has to be tackled on this PR though.

PR #787, opened 2026-07-29
[BugFix] Support TLS on the operator's MySQL connection to FE

yandongxiao:
# Description

The operator opens its MySQL connection to FE (used for `SHOW COMPUTE NODES`,
`ALTER SYSTEM DROP COMPUTE NODE`, `DROP WAREHOUSE`) in **plaintext only**, with no switch to enable
TLS. When a user sets `ssl_force_secure_transport = true` in `fe.conf`, FE rejects every operator
connection with `Error 5205 (HY000): Connections using insecure transport are prohibited` and logs
a WARN for each attempt — roughly 10–14 failures per FE per minute in a real deployment, burying
real errors in log noise. There is exactly one MySQL connection site in the repo
(`pkg/subcontrollers/cn/cn_mysql.go`, two `sql.Open` call sites) and its DSN ends in a bare `/`.

**Fix:** append a `tls=` parameter to that DSN, controlled by one new global operator flag
`--fe-ssl-mode` (Helm value `starrocksOperator.feSslMode`), with values `DISABLED` / `PREFERRED` /
`REQUIRED`, default `PREFERRED`.

| `--fe-ssl-mode` | DSN suffix | Semantics |
|---|---|---|
| `DISABLED` | none | plaintext; byte-for-byte the behavior that predates this change |
| `PREFERRED` (**default**) | `?tls=preferred` | encrypt when FE advertises `CLIENT_SSL`, stay plaintext otherwise |
| `REQUIRED` | `?tls=skip-verify` | always encrypt; fail when FE does not support SSL |
| `VERIFY_CA` / `VERIFY_IDENTITY` | — | recognized but **rejected at startup**, with an error explaining why |

Values are matched case-insensitively; an unrecognized or empty value makes the operator **fail to
start** rather than silently fall back to plaintext. The server certificate is never verified in
any mode — FE's own documentation generates a self-signed keystore whose certificate name cannot
match the Kubernetes service DNS name, so `VERIFY_CA` / `VERIFY_IDENTITY` would be unusable in the
documented deployment; the two names are reserved so certificate verification can be added later
without changing the parameter.

**Why `PREFERRED` is a safe default:** an FE without a keystore does not advertise `CLIENT_SSL`, so
the driver takes its `AllowFallbackToPlaintext` branch and behaves exactly like today. This was
measured against a live FE before fixing the default (no `tls` param → success, `?tls=preferred` →
success, `?tls=skip-verify` → `TLS requested but server does not support TLS`).

**No new dependency, no CRD change, no RBAC change, no certificate mounting.**
`go-sql-driver/mysql` is already vendored and `preferred` / `skip-verify` are its native values.

Changes:

- `pkg/subcontrollers/cn/cn_mysql_ssl.go` (new) — the mode constants, `ValidateSSLMode`,
  `sslModeDSNSuffix`.
- `pkg/subcontrollers/cn/cn_mysql.go` — the two duplicated DSN literals collapsed into one `dsn()`
  method, so a future change cannot make one query path negotiate TLS while the other stays
  plaintext.
- `cmd/config/config.go`, `cmd/main.go` — the flag, validated immediately after `flag.Parse()` so a
  bad value fails at startup instead of at the first reconcile.
- Operator subchart — `starrocksOperator.feSslMode: PREFERRED` plus a `{{- if }}`-gated
  `--fe-ssl-mode` arg (following the `--dns-domain-suffix` precedent): setting the value to `""` or
  `null` omits the whole arg and falls back to the binary default, instead of rendering
  `--fe-ssl-mode=` and killing the operator on start.
- `deploy/operator.yaml` and the parent `kube-starrocks/values.yaml` — regenerated
  (`scripts/operator.sh`, `scripts/create-parent-chart-values.sh`).
- `doc/connect_to_ssl_enabled_fe_howto.md` (new) — how the flag works, why certificates are not
  verified, how to choose a mode, FE-side keystore traps (PKCS12 `-keypass` silently ignored;
  loopback connections exempt from `ssl_force_secure_transport`), and both Helm and raw-manifest
  configuration. Indexed in `doc/README.md`.

**Upgrade behavior:** nothing in the StatefulSet / Pod rendering path is touched, so FE / BE / CN
pods are not restarted — only the operator's own Deployment rolls. The only behavioral change is
that an FE which already advertises SSL will get an encrypted operator connection; set
`feSslMode: DISABLED` to return to the exact previous behavior.

Verification on this branch:

- 24 new/extended unit subtests across `TestSSLModeDSNSuffix`, `TestValidateSSLMode`,
  `TestSQLExecutorDSN`, `TestNewSQLExecutor`; `go test ./pkg/...` green.
- `helm lint` clean; render checked end to end through the parent chart: default renders
  `--fe-ssl-mode=PREFERRED`, `--set operator.starrocksOperator.feSslMode=DISABLED` passes through,
  an empty value omits the arg.
- `make generate` / `make manifests` produce no diff; `golangci-lint run` (v2.2.2, the CI-pinned
  version): 0 issues.
- End-to-end behavior was verified on a kind cluster against FE 4.x with a real SSL keystore:
  with SSL enforced and mode `DISABLED` the operator logs error 5205 matching FE-side WARNs
  one-to-one; with the default `PREFERRED` on the same FE, `SHOW COMPUTE NODES` and
  `ALTER SYSTEM DROP COMPUTE NODE` succeed (which on a plaintext-rejecting FE can only mean the
  connection was upgraded to TLS); `VERIFY_CA` and an invalid value each put the new operator pod
  into CrashLoopBackOff with the intended message while the old pod keeps running. A `helm upgrade`
  across this change left FE/CN StatefulSets untouched (same UID / restartCount /
  `controller-revision-hash` / `resourceVersion`).

# Related Issue(s)

N/A — reported by an enterprise customer running the operator against an FE with
`ssl_force_secure_transport = true` (operator v1.11.x has the defect).

# Checklist

- [x] run `make generate` to generate the code. — no diff (no API-type change).
- [x] run `make lint` to check the code style. — golangci-lint v2.2.2: 0 issues.
- [x] run `make test` to run UT. — all packages ok, including the 24 new subtests in
  `pkg/subcontrollers/cn`.
- [x] run `make manifests` to update the yaml files of CRD. — no CRD diff (this is a global
  operator flag by design, not a CRD field).

- [x] make sure you have updated the `values.yaml` file of starrocks chart. — the value belongs to
  the **operator** subchart (`kube-starrocks/charts/operator/values.yaml`,
  `starrocksOperator.feSslMode: PREFERRED`); the `starrocks` subchart is deliberately untouched —
  this flag configures the operator process, not the data plane.
- [x] In `scripts` directory, run `bash create-parent-chart-values.sh` to update the values.yaml
  file of the parent chart (kube-starrocks chart). — regenerated, committed.

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -8,6 +8,12 @@ var (
 
 	// VolumeNameWithHash decides whether adding a hash to the volume name
 	VolumeNameWithHash bool
+
+	// FeSslMode decides whether the operator's MySQL connection to FE negotiates TLS. Valid values
+	// are DISABLED, PREFERRED and REQUIRED — the MySQL client's --ssl-mode vocabulary, matched
+	// case-insensitively. It is validated once at startup (see cn.ValidateSSLMode), applies to the
+	// whole operator process, and requires an operator restart to change.
+	FeSslMode string
 )
 
 func GetServiceDomainSuffix() string {
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -29,6 +29,7 @@ import (
 	srapi "github.com/StarRocks/starrocks-kubernetes-operator/pkg/apis/starrocks/v1"
 	"github.com/StarRocks/starrocks-kubernetes-operator/pkg/controllers"
 	"github.com/StarRocks/starrocks-kubernetes-operator/pkg/k8sutils"
+	"github.com/StarRocks/starrocks-kubernetes-operator/pkg/subcontrollers/cn"
 )
 
 var (
@@ -50,6 +51,11 @@ func main() {
 	flag.StringVar(&config.DNSDomainSuffix, "dns-domain-suffix", "cluster.local", "The suffix of the dns domain in k8s")
 	flag.BoolVar(&config.VolumeNameWithHash, "volume-name-with-hash", true, "Add a hash to the volume name")
 	flag.StringVar(&_denyList, "deny-list", "", "Comma-separated list of namespaces to exclude from reconciliation")
+	flag.StringVar(&config.FeSslMode, "fe-ssl-mode", cn.SSLModePreferred,
+		"How the operator's MySQL connection to FE negotiates TLS. DISABLED keeps it plaintext; "+
+			"PREFERRED encrypts when FE advertises SSL support and stays plaintext otherwise; "+
+			"REQUIRED always encrypts and fails when FE does not support SSL. Case-insensitive. "+
+			"The server certificate is never verified. Requires an operator restart to change.")
 
 	// Set up logger.
 	opts := zap.Options{}
@@ -57,6 +63,12 @@ func main() {
 	flag.Parse()
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 	logger := ctrl.Log.WithName("main")
+
+	// Fail fast: an unusable ssl mode would otherwise stay silent until the first FE connection.
+	if err := cn.ValidateSSLMode(config.FeSslMode); err != nil {
+		logger.Error(err, "invalid --fe-ssl-mode")
+		os.Exit(1)
+	}
 
 	// Register CRD to SchemeBuilder
 	srapi.Register()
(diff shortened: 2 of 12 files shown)


kevincai:
@codex review

kevincai (approved):

PR #791, opened 2026-08-05
[Feature] Support custom labels for operator Deployment and pod

### LINKED ISSUE #790: Support custom labels on the operator Deployment and pod
## Describe the current behavior

The operator Helm chart only adds its built-in `app` and `version` labels to the operator Deployment and pod. It supports custom annotations, but it does not provide a value for adding custom labels.

Users that require ownership, cost-allocation, policy, or workload-integration labels must maintain a downstream copy of the chart to modify `templates/deployment.yaml`.

## Describe the enhancement

Add an optional `starrocksOperator.labels` map to the operator chart. When configured, render these labels on:

- the operator Deployment metadata;
- the operator pod-template metadata.

Keep the Deployment selector unchanged. Custom organizational labels are mutable and should not become part of the immutable `spec.selector`.

Example:

```yaml
starrocksOperator:
  labels:
    team: data-platform
    environment: production
```

The default should remain `{}` so existing rendered manifests do not change.

## Additional context

This would remove the need for downstream chart forks whose only template modification is custom operator labels. The chart already exposes `starrocksOperator.annotations`; providing the equivalent label capability would make its metadata configuration more complete.

TiansuYu:
## Description

Add an optional `starrocksOperator.labels` value to the operator Helm chart and render it on the operator Deployment and pod template.

Custom labels are intentionally not added to `spec.selector`. This keeps the selector stable and allows organizational labels to change without requiring the Deployment to be recreated.

The parent `kube-starrocks` values file has been regenerated with `scripts/create-parent-chart-values.sh`.

## Related Issue(s)

Closes #790

## Validation

- `helm lint helm-charts/charts/kube-starrocks/charts/operator`
- `helm lint helm-charts/charts/kube-starrocks`
- Rendered the operator chart with default values and confirmed its output is unchanged.
- Rendered the operator and parent charts with custom labels and confirmed they appear on Deployment and pod metadata but not on `spec.selector`.

## Checklist

Operator code was not changed, so code generation, CRD generation, and Go unit tests are not applicable.

For the Helm chart:

- [x] Updated the operator child-chart `values.yaml`.
- [x] Ran `scripts/create-parent-chart-values.sh` to regenerate the parent `kube-starrocks` values.
- [x] Ran Helm lint and template validation for the operator and parent charts.

### DIFF
--- a/helm-charts/charts/kube-starrocks/charts/operator/templates/deployment.yaml
+++ b/helm-charts/charts/kube-starrocks/charts/operator/templates/deployment.yaml
@@ -6,6 +6,9 @@ metadata:
   namespace: {{ template "operator.namespace" . }}
   labels:
     app: {{ template "operator.name" . }}-operator
+    {{- with .Values.starrocksOperator.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
   {{- if .Values.starrocksOperator.annotations }}
   annotations:
     {{- toYaml .Values.starrocksOperator.annotations | nindent 4 }}
@@ -25,6 +28,9 @@ spec:
       labels:
         app: {{ template "operator.name" . }}-operator
         version: {{ $.Chart.Version }}
+        {{- with .Values.starrocksOperator.labels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
     spec:
       automountServiceAccountToken: true
       {{- if .Values.starrocksOperator.imagePullSecrets }}
--- a/helm-charts/charts/kube-starrocks/charts/operator/templates/deployment.yaml
+++ b/helm-charts/charts/kube-starrocks/charts/operator/templates/deployment.yaml
@@ -1,8 +1,16 @@
 {{- if .Values.starrocksOperator.enabled }}
 {{- $operatorLabels := .Values.starrocksOperator.labels | default dict }}
-{{- range $reservedLabel := list "app" "version" }}
+{{- if not (kindIs "map" $operatorLabels) }}
+{{- fail "starrocksOperator.labels must be a map" }}
+{{- end }}
+{{- range $reservedLabel := list "app" "version" "app.kubernetes.io/managed-by" }}
 {{- if hasKey $operatorLabels $reservedLabel }}
 {{- fail (printf "starrocksOperator.labels cannot override reserved label %q" $reservedLabel) }}
+{{- end }}
+{{- end }}
+{{- range $key, $value := $operatorLabels }}
+{{- if or (kindIs "map" $value) (kindIs "slice" $value) (kindIs "invalid" $value) }}
+{{- fail (printf "starrocksOperator.labels value for %q must be a string, number, or boolean" $key) }}
 {{- end }}
 {{- end }}
 apiVersion: apps/v1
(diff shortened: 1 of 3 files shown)


kevincai (dismissed):

TiansuYu:
@kevincai I have fixed a bug spotted by Copilot, do you think you can approve this and we get this merged?

kevincai (dismissed):

yandongxiao (commented):
Thanks for this. The core design decision is the right one, and it holds up under testing: I ran the chart on a kind cluster and confirmed that custom labels land on the Deployment and the pod template, `spec.selector` stays untouched, and a `helm upgrade` that adds — or later removes — labels on an existing release is a clean in-place update, with no immutable-selector failure and the original ReplicaSet reused once the labels are removed again. Default rendering is byte-identical to `main`.

Two things I ran into while testing, one of which I'd like to see fixed before merge.

### 1. Non-string label values render an invalid manifest

`toYaml` emits each value with its YAML type, so any label set without `--set-string` renders as a scalar the API server cannot decode:

```console
$ helm template ... --set starrocksOperator.labels.rev=123
  labels:
    app: kube-starrocks-operator
    rev: 123          # wanted "123"
```

```console
$ helm upgrade ... --set starrocksOperator.labels.rev=123
Error: UPGRADE FAILED: unable to decode "": json: cannot unmarshal number into Go struct field ObjectMeta.metadata.labels of type string
```

Booleans behave the same way (`prod: true`). `--set-string` works around it, but the error names neither a file nor a values key, so it is hard to trace back to `starrocksOperator.labels`. Revision numbers, ticket ids and cost-center codes are exactly the kind of values people put in labels, so I think it's worth quoting them in the chart:

```yaml
    {{- range $k, $v := $operatorLabels }}
    {{ $k }}: {{ $v | quote }}
    {{- end }}
```

and the same at `nindent 8`'s indentation for the pod template. I tried this on your branch: default rendering stays byte-identical, `rev=123` / `prod=true` then render as `"123"` / `"true"`, and the result applies cleanly.

### 2. `app.kubernetes.io/managed-by` silently diverges between Deployment and pod (nit)

Helm sets `app.kubernetes.io/managed-by: Helm` on release resources itself, so a user-supplied value for that key survives on the pod template but is overwritten on the Deployment:

```console
$ helm upgrade ... --set 'starrocksOperator.labels.app\.kubernetes\.io/managed-by=my-gitops'
deploy.metadata.labels[managed-by] : Helm
pod-template.labels[managed-by]    : my-gitops
```

Harmless in itself, but it makes ownership queries inconsistent between the Deployment and its pods. Adding the key to the reserved list alongside `app` / `version` would turn a silent surprise into an explicit error. Entirely your call — feel free to treat this as a nit.

One more thing worth a line in the PR description: because the labels also go on the pod template, setting or changing them rolls the operator pod once. That is expected and does not touch FE/BE/CN, but it's good for users to know up front.

yandongxiao (changes requested):

TiansuYu:
@yandongxiao Thanks for the detailed review, I have addressed the concerns you have mentioned. For 2rd point, I have rejected this particular label from acceptable key-value pairs, should be managed by deployment controller any way.

kevincai (approved):

yandongxiao (approved):

Issue #617, opened 2025-01-24
Can't change storage for existing cluster


dorsegal:
## Describe the bug

I want to add separate spill volume for CN but operator does not support changing PVC and PVC templates.
It will fail to manage increasing volume.

```
is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'persistentVolumeClaimRetenti │
│ onPolicy' and 'minReadySeconds' are forbidden"}   
```

## Expected behavior

Operator should handle storage changes like other operators. See Strimzi for Kafka.

## Please complete the following information

- Operator Version: v1.9.10
- Chart Name: kube-starrocks
- Chart Version 1.9.10

yandongxiao:
I think this is a known issue. Because the internal implementation of Operator relies on Statefulset and its `volumeClaimTemplates` field. Once this field is created, it cannot be changed. Currently, one workaround you can use is to uninstall the Helm Chart and then reinstall the Chart. Since PVC will not be deleted during the uninstall process, the data in the original cluster will be reused.

dorsegal:
My workaround was deleting Statefulset of CN using orphan. This keeps everything and the operator created new Statefullset with the desired state. Rolling restart to update pods.

Should I leave this open? Will this be implemented?

shyundev:
@varun-2108 the order matters: change the StarRocksCluster first, then delete the StatefulSet with `--cascade=orphan`. The operator recreates the StatefulSet on the next reconcile from what the CR says at that moment, so deleting first hands it the old spec. #800 documents this in `doc/mount_ephemeral_volumes_howto.md`, and adds an `ephemeral` flag under which `storageSize` becomes editable for cache and spill volumes.

Issue #419, opened 2024-01-24
StarRocks in Kubernetes should allow to use HostPath for CN nodes' cache


alberttwong:
StarRocks in Kubernetes should allow to use HostPath for CN nodes' cache. Using PVCs isn't good hear, especially if I wanna make use of NVMe EC2 instance types. Right now I'm running a cluster of m6id.2xlarge and I had to deploy the Local Volume Provisioner in K8s which isn't trivial, and it also breaks auto-scaling.

@nferrario

yandongxiao:
Could you use 'emptyDir' as CN nodes' cache ? This is the default volumes CN nodes use.

nferrario:
Sadly no. Empty dir uses the root volume of Kubernetes, which in EC2 it would be an EBS backed volume. My goal is to leverage Instance Storage, which offers much lower latency and higher throughput but is ephemeral.

Issue #121, opened 2023-05-26
[Feature] Support resize PVC or adding more PVC to BE


dengliu:

kevincai:
PVC resizing is out of operator's control. due to limitation of statefulset in current implementation, add more PVCs are not allowed after initial installation.

nferrario:
The operator should allow for updating the PVC request size. If the underlying Storage Class supports expansion (like EBS), the volume will expand automatically and transparently.
I just had to manually edit the PVC to increase storage and the sad side-effect is that now my StarRocks cluster YAML is out of sync with the PVCs' new size

dorsegal:
+1
The ability to increase pvc size is crucial for operator functionality. 
I know for fact that some operators handle it (ie. strimzi kafka)

OliveBZH:
+1 
This resizing is critical for us as well

hrvatskibogmars:
+1, manually updated the PVC but left operator config on older value so its not out of sync on operator side....

patrickmanzo:
+1 waiting this feature :)

PR #800, opened 2026-09-07
[Enhancement] Support generic ephemeral volumes for cache and spill data

shyundev:
# Description

A CN in a shared-data cluster keeps its data cache on local disk, and BE and CN spill there. Both
want a dedicated disk of a declared size, one per pod, holding data that can be rebuilt. Today a
`storageVolumes` entry with a storage class becomes a `volumeClaimTemplate` of the StatefulSet,
and that claim outlives the pod on purpose. This PR adds one flag to `StorageVolume`:

```yaml
storageVolumes:
  - name: cn-data
    storageClassName: local-nvme
    storageSize: 500Gi
    mountPath: /opt/starrocks/cn/storage
    ephemeral: true
```

The volume keeps its storage class and size, but the claim is declared in the pod spec as a
[generic ephemeral volume](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes)
instead of in `volumeClaimTemplates`: Kubernetes creates the PersistentVolumeClaim with the pod,
names it `<pod>-<volume>`, deletes it with the pod, and a replacement pod provisions a new volume
where it is scheduled. A reserved `storageClassName` would not fit here: `emptyDir`, `hostPath`
and `csi` (#786) all do without a claim, while an ephemeral volume still needs a real storage
class to provision from.

Observed on a three-worker cluster (Kubernetes v1.30.4) running this branch, with two CRs that
differ only in `ephemeral`, on a local-path `WaitForFirstConsumer` class:

| Scenario | `volumeClaimTemplates` (today) | `ephemeral: true` |
| --- | --- | --- |
| The pod's node is cordoned and the pod deleted (#419) | stays `Pending` | `Running` on another node, on a new empty volume |
| `storageSize` is raised from 1Gi to 2Gi (#121, #617) | cluster `failed`, `Forbidden` | rolling update, the new pod claims 2Gi |

Row 1: a local PersistentVolume carries node affinity, so the replacement pod can only bind to
the node that is gone. `persistentVolumeClaimRetentionPolicy` does not help: it deletes claims
when the StatefulSet is deleted or scaled down, not when a pod is rescheduled.

Row 2: `volumeClaimTemplates` is immutable, so the update fails and `status.reason` reports:

```
error from CN controller: failed to apply merge patch: StatefulSet.apps "pinned-cn" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
```

An ephemeral claim sits in `spec.template`, which is mutable. This is not in-place expansion: the
change rolls the pods and each new pod claims the new size.

What this PR changes:

- `pod.MountStorageVolumes` renders an `ephemeral` volume through the new `MountEphemeralVolume`
  and `statefulset.PVCList` skips it, so the claim is declared once. Both build it with
  `pod.PersistentVolumeClaimSpec`, lifted out of `PVCList`.
- Validation rejects `ephemeral` together with `emptyDir`, `hostPath` or `csi`, given as
  `storageClassName` (case insensitive, like the reserved values themselves) or as a block:
  those render no claim, so the flag would be dropped silently.
- Helm: `storageSpec.storageEphemeral` for the data volumes and `storageSpec.spillEphemeral` for
  the spill volume, on `starrocksCnSpec` and `starrocksBeSpec`.
- Docs: `doc/mount_ephemeral_volumes_howto.md`, linked from `doc/README.md`.

The operator change is under a hundred lines; the rest is tests, chart values, the how-to and
generated CRD yaml.

`ephemeral` defaults to false, so no existing spec renders differently. Like `csi`, it reaches
all four component specs and `StarRocksWarehouse` through the shared `StorageVolume` type.
Generic ephemeral volumes are GA since Kubernetes 1.23, the chart's `kubeVersion` floor. Turning
`ephemeral` on or off on a deployed volume hits the same `Forbidden` as any other
`volumeClaimTemplates` change; the how-to gives the `--cascade=orphan` path out.

To see it on a cluster, apply the sample CR from the how-to and check where each claim comes from:

```bash
# expected: cn-log
kubectl get statefulset starrockscluster-sample-cn -o jsonpath='{.spec.volumeClaimTemplates[*].metadata.name}'
# expected: starrockscluster-sample-cn-0-cn-data owned by Pod, cn-log-starrockscluster-sample-cn-0 by <none>
kubectl get pvc -o custom-columns=NAME:.metadata.name,OWNER:.metadata.ownerReferences[*].kind
```

# Related Issue(s)

- #121 (open) and #617 (open): `storageSize` cannot be changed once the cluster exists. With
  `ephemeral` it can, for the volumes that carry the flag, by re-provisioning rather than
  expanding.
- #419: a CN cache on NVMe instance storage; the reporter fell back to a local volume provisioner,
  which "breaks auto-scaling". Closed by #451 with `hostPath`.

# Checklist

- [x] run `make lint` to check the code style.

  file of starrocks chart.
  chart( kube-starrocks chart).

### DIFF
--- a/config/crd/bases/starrocks.com_starrocksclusters.yaml
+++ b/config/crd/bases/starrocks.com_starrocksclusters.yaml
@@ -4218,6 +4218,14 @@ spec:
                           required:
                           - driver
                           type: object
+                        ephemeral:
+                          description: |-
+                            Ephemeral makes this volume a generic ephemeral volume: its PersistentVolumeClaim is declared
+                            in the pod spec, owned by the pod and deleted with it, so a replacement pod provisions a new
+                            volume. StorageClassName and StorageSize still apply, but the data does not outlive the pod,
+                            so use it for a cache or a spill directory. It can not be set with emptyDir, hostPath or csi.
+                            More info: https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes
+                          type: boolean
                         hostPath:
                           description: |-
                             HostPath Represents a host path mapped into a pod.
@@ -9330,6 +9338,14 @@ spec:
                           required:
                           - driver
                           type: object
+                        ephemeral:
+                          description: |-
+                            Ephemeral makes this volume a generic ephemeral volume: its PersistentVolumeClaim is declared
+                            in the pod spec, owned by the pod and deleted with it, so a replacement pod provisions a new
+                            volume. StorageClassName and StorageSize still apply, but the data does not outlive the pod,
+                            so use it for a cache or a spill directory. It can not be set with emptyDir, hostPath or csi.
+                            More info: https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes
+                          type: boolean
                         hostPath:
                           description: |-
                             HostPath Represents a host path mapped into a pod.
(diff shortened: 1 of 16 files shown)


shyundev:
@kevincai this is my first PR in this repo, so the workflow runs are waiting for approval. Could you enable them?
@yandongxiao this adds an `ephemeral` flag to `StorageVolume`, next to the `csi` field from #786. Could you take a look when you have time?

Issue #738, opened 2026-02-03
Enhance Rolling Upgrade Behavior


AdheipSingh:
## Describe the current behavior

During a rolling upgrade, the operator checks FE readiness before updating BE/CN components using `CheckFEReady()` function (`pkg/subcontrollers/fe/fe_controller.go:237-259`). This function only verifies   
    that **at least one FE pod has an endpoint**:                                                                                                                                                                
                                                                                                                                                                                                                 
   ```go                                                                                                                                                                                                         
   func CheckFEReady(...) bool {                                                                                                                                                                                 
       for _, sub := range endpoints.Subsets {                                                                                                                                                                   
           if len(sub.Addresses) > 0 {                                                                                                                                                                           
               return true  // Returns true if ANY endpoint exists                                                                                                                                               
           }                                                                                                                                                                                                     
       }                                                                                                                                                                                                         
       return false                                                                                                                                                                                              
   }                                                                                                                                                                                                             
   ```                                                                                                                                                                                                           
                                                                                                                                                                                                                 
   This means:                                                                                                                                                                                                   
   - BE/CN updates can start as soon as **one FE pod** is ready                                                                                                                                                  
   - The operator does **not** wait for the FE StatefulSet to be fully rolled out (all pods at new revision)                                                                                                     
   - During FE rolling upgrade, BE/CN updates may proceed while FE is still partially updated                                                                                                                    
                                                                                                                                                                                                                 
   ## Potential Risk                                                                                                                                                                                             
                                                                                                                                                                                                                 
   If a bad image or configuration is rolled out to FE:                                                                                                                                                          
   1. FE pod-0 gets updated to bad version                                                                                                                                                                       
   2. FE pod-0 becomes ready (passes readiness probe)                                                                                                                                                            
   3. `CheckFEReady()` returns true (one endpoint exists)                                                                                                                                                        
   4. BE/CN StatefulSets start updating to new version                                                                                                                                                           
   5. FE pod-0 later fails or exhibits bad behavior                                                                                                                                                              
   6. Now both FE and BE/CN are in a bad state                                                                                                                                                                   
                                                                                                                                                                                                                 
   In contrast, if we waited for **full FE StatefulSet revision completion**:                                                                                                                                    
   1. FE pod-0 gets updated to bad version                                                                                                                                                                       
   2. FE pod-0 fails or gets stuck                                                                                                                                                                               
[...]

PR #707, opened 2025-10-09
[Bugfix] Implement upgrade-aware controller ordering for FE/BEs/CNs

jmjm15x:
# Description

Add upgrade sequencing control to the StarRocks Kubernetes Operator to ensure proper component ordering during both initial deployments and upgrades. Previously, the operator always used FE-first ordering, which is correct for initial deployments but incorrect for upgrades. According to StarRocks guidelines:

- **Initial Deployment:** FE → BEs/CNs (FE must be leader before workers join)
- **Cluster Upgrades:** BEs/CNs → FE (data nodes upgraded before metadata nodes)

From [official documentation](https://docs.starrocks.io/docs/deployment/upgrade/#upgrade-procedure):

> **Upgrade procedure**
> By design, BEs and CNs are backward compatible with the FEs. Therefore, you need to upgrade BEs and CNs first and then FEs to allow your cluster to run properly while being upgraded. Upgrading them in an inverted order may lead to incompatibility between FEs and BEs/CNs, and thereby cause the service to crash.

## Solution

Implemented a comprehensive upgrade detection and sequencing mechanism with robust component readiness validation to prevent premature progression between components.

#### 1. **Upgrade Detection (`isUpgrade()`)**
Detects upgrade scenarios by checking if StatefulSets exist with pending changes, compares `Generation` vs `ObservedGeneration` to detect spec changes.

**Why this approach?**
- Simple and reliable, uses Kubernetes native generation tracking
- Works for any spec change (images, resources, configs)
- Handles transient states, correctly identifies upgrade during rollout progress

#### 2. **Controller Ordering (`getControllersInOrder()`)**
Dynamically switches controller execution order based on deployment scenario:
- **Upgrade scenario:** `[be, cn, fe, feproxy]` (BE-first ordering)
- **Initial deployment:** `[fe, be, cn, feproxy]` (FE-first ordering)

#### 3. **Component Readiness Validation (`isComponentReady()`)**
Multi-layer validation to prevent premature component progression by avoiding race conditions, ensuring rollout stability, and providing enhanced logging for debugging.

### Logic Flow

Implements waiting logic directly in the reconciliation loop:

```
Reconcile() called
    ↓
Get controllers in order based on isUpgrade()
    ↓
For each controller in order:
    ↓
    ├─ If upgrade && feController
    │   └─ Check BE/CN ready? → If NO, wait and requeue
    │
    ├─ Sync controller (create/update resources)
    │
    ├─ If initial && feController  
[...]

PR #801, opened 2026-09-14
[BugFix] waitForFullRollout has no effect when FE and CN images change together

carlosescura:
# Description

`waitForFullRollout` (added in #739) is a no-op in the exact case it was designed for: an upgrade where the FE and BE/CN images change in the same `StarRocksCluster` update. BE/CN roll concurrently with FE instead of waiting.

## Root cause

`CheckFEFullyRolledOut` reads only `sts.Status` and never compares it against `metadata.Generation`.

Within a single reconcile, `feController.SyncCluster` updates the FE StatefulSet spec, then `cnController.SyncCluster` (or `beController.SyncCluster`) runs immediately after and calls `CheckFEFullyRolledOut`. At that moment the StatefulSet controller has not yet reconciled the new generation, so `Status` still describes the previous one:

- `ReadyReplicas` still equals `Spec.Replicas` — the old pods are all still running and ready
- `UpdateRevision` still equals `CurrentRevision` — the new revision hash has not been computed yet

Both checks pass, the function returns `true`, and BE/CN is synced in the same reconcile that updated FE. Both StatefulSets then roll at the same time.

## Observed behaviour

StarRocks 4.0.9 → 4.1.3 on a shared-data cluster (3 FE, 2 CN), operator v1.11.7, CRD v1.11.7, with `spec.waitForFullRollout: true` confirmed present on the live CR:

- An FE pod and a CN pod were replaced **4 seconds apart**
- `"not fully rolled out, skipping CN sync"` appears **0 times** in the operator log for the entire rollout

The same behaviour occurred on a staging cluster on the same upgrade path.

## Fix

Return `false` when the StatefulSet controller has not yet observed the current spec — the standard guard against acting on stale `Status`.

Both `be_controller.go` and `cn_controller.go` call `CheckFEFullyRolledOut`, so this covers BE and CN.

`CheckFEReady` (the `waitForFullRollout: false` path) is deliberately unchanged: it reads FE Service endpoints rather than StatefulSet status, and its "any FE pod is ready" semantics are intentional.

## Tests

Two cases added to `TestCheckFEFullyRolledOut`:

- **`statefulset spec updated but status not yet observed`** — reproduces the bug. Fails on `main` with `CheckFEFullyRolledOut() = true, want false`; passes with the fix.
- **`statefulset fully rolled out with observed generation`** — confirms non-zero matching generations still report rolled out.

All five pre-existing cases still pass; they use zero-valued generations, so the guard is a no-op for them.

# Related Issue(s)

Implements the behaviour intended by #738 / #739.

Independent of #707 (upgrade-aware controller ordering) and #792 (`OnDelete` strategy in rollout status) — no overlap with either.

# Checklist

For operator:

- [ ] run `make generate` — not needed, no API type changes
- [x] run `make lint` — golangci-lint v2.2.2 (repo-pinned), 0 issues
- [x] run `make test` — `./pkg/subcontrollers/...` all pass. The one failure in `./pkg/...` is `TestSetupClusterReconciler`, a pre-existing local envtest issue (missing `kubebuilder/bin/etcd`) that reproduces identically on unmodified `main`.
- [ ] run `make manifests` — not needed, no CRD changes

For helm chart: not applicable. This is an operator-only fix; `waitForFullRollout` is already exposed in the chart via #771, and no values or CRD change is required.


carlosescura:
@kevincai @yandongxiao 

This is not a critical bug, but might be worth checking it soon as it fixes a false feeling of "security" when rolling out updates.

